PR #45, opened 2018-04-09
adds CloudFront cache invalidation to deployments

tgross:
CloudFront transmits the Cache-Control headers from the origin but does not respect them themselves in the special case of no-cache if the TTL minimum is >0. In this case CloudFront uses the distribution's TTL as one would expect of a CDN.

We can't cache-bust the index document by changing the name the way we do the rest of our dashboard's files, so this forces an invalidation in CloudFront via the AWS CLI.

For [DEN-4334](https://jira.density.io/browse/DEN-4334)


tgross:
Before we can merge this I think need to update the CircleCI IAM user's permissions. But I've tested the cache invalidation process manually by making sure I got a cache hit, sending the invalidation, waiting for the invalidation to complete, and then making sure I got a cache miss on the next request.

```
$ curl -vso /dev/null https://dashboard.density.rodeo
...
< HTTP/1.1 200 OK
< Content-Type: text/html
< Content-Length: 556
< Connection: keep-alive
< Date: Mon, 09 Apr 2018 12:42:09 GMT
< Cache-Control: no-cache
< Last-Modified: Thu, 22 Mar 2018 19:32:48 GMT
< ETag: "871047b4ede04bd382d8ec9992431a7c"
< Server: AmazonS3
< X-Cache: Miss from cloudfront
< Via: 1.1 87ff2ecc6cfa7d45348d40b23644c2a5.cloudfront.net (CloudFront)
< X-Amz-Cf-Id: 2ZqzqRofOj9tb4ODPDfUzVnMtkIww_3Kr2jarpQ8TNQnEM0MkxSJlQ==


$ curl -vso /dev/null https://dashboard.density.rodeo
...
< HTTP/1.1 200 OK
< Content-Type: text/html
< Content-Length: 556
< Connection: keep-alive
< Date: Mon, 09 Apr 2018 12:42:09 GMT
< Cache-Control: no-cache
< Last-Modified: Thu, 22 Mar 2018 19:32:48 GMT
< ETag: "871047b4ede04bd382d8ec9992431a7c"
< Server: AmazonS3
< X-Cache: RefreshHit from cloudfront
< Via: 1.1 c40746e561ae602a5c205b30ae66f603.cloudfront.net (CloudFront)
< X-Amz-Cf-Id: LO3XHauLMQjxkmK733VQrWnvH0AxW4QVnC0IIDdwQ__fRqIJ7NvQZg==

$ aws cloudfront create-invalidation --distribution-id  EJGYK3LXP76SR --paths "/"
{
    "Invalidation": {
        "Status": "InProgress",
        "InvalidationBatch": {
            "Paths": {
                "Items": [
                    "/"
                ],
                "Quantity": 1
            },
            "CallerReference": "cli-1523277750-175307"
        },
        "Id": "I238WDOLR208DS",
        "CreateTime": "2018-04-09T12:42:29.590Z"
    },
    "Location": "https://cloudfront.amazonaws.com/2017-03-25/distribution/EJGYK3LXP76SR/invalidation/I238WDOLR208DS"
}

$ aws cloudfront create-invalidation --distribution-id  EJGYK3LXP76SR --paths "/"
...
< HTTP/1.1 200 OK
< Content-Type: text/html
< Content-Length: 556
< Connection: keep-alive
< Date: Mon, 09 Apr 2018 12:44:12 GMT
< Cache-Control: no-cache
< Last-Modified: Thu, 22 Mar 2018 19:32:48 GMT
< ETag: "871047b4ede04bd382d8ec9992431a7c"
< Server: AmazonS3
< X-Cache: Miss from cloudfront
< Via: 1.1 44acf4d0ae23dab6e6e1414b314e9296.cloudfront.net (CloudFront)
< X-Amz-Cf-Id: duEf-KI7aUJpbspunNqiCIRG2UzWjL0HVUsVOqlO6fI-fTHqadSP1Q==
```

beweinreich:
Ugh it sucks that we can’t cache bust the index page. All the other cache busting stuff is useless if you can’t bust the index page to point to all the new files. 

So it seems like we’re beholden to the slow cloudfront invalidation time. 

I wonder if you can swap out the “root document” in an s3 bucket website. Change it from index.html to index-<timestamp>.html  Never seen anyone do that before so it might be a rabbit hole.

1egoman:
@beweinreich That could work maybe for the staging site, but wouldn't effect the preview builds, because each preview version of the site is hosted at `/<commit hash>/index.html` within that bucket.

tgross:
> Ugh it sucks that we can’t cache bust the index page. All the other cache busting stuff is useless if you can’t bust the index page to point to all the new files.

Not really. We don't control the order of upload to S3 so if we had instantaneous cutover to the new index page we'd potentially have a bunch of 404s on content that the new index page points to but that haven't been uploaded. Relying on the cache simplifies the deployment quite a bit.

> So it seems like we’re beholden to the slow cloudfront invalidation time.

We could switch CDNs to one that's faster to invalidate (ex. Fastly) but that will probably cost us more both in $ and time to switch.

> That could work maybe for the staging site, but wouldn't effect the preview builds, because each preview version of the site is hosted at /preview/<commit hash>/index.html within that bucket.

Sure but we don't need to bust cache for the preview versions, because each commit has it's own SHA and that busts cache automatically.

1egoman:
> Sure but we don't need to bust cache for the preview versions, because each commit has it's own SHA and that busts cache automatically.

You'd think that, but I think something in the pipeline defaults to returning `/index.html` instead of the proper `/preview/<hash>/index.html`(Maybe it's the browser? I haven't really done any sort of digging into it). Ben (and I) have both run into caching issues with the preview builds right after completion, but after force-reloading the browser / using incognito the proper index.html is returned.

tgross:
> Maybe it's the browser? I haven't really done any sort of digging into it

Ok, well this is where the investigation should have started instead of changing things in the infrastructure.

PR #50, opened 2018-04-24
Log to telemetry 

1egoman:
Send logs to the telemetry service for that relate to sockets events.

- Adds a new `logger` helper that can be called anywhere in the application to display logs.
  - This is still behind the scenes still using `debug`, `logger` is just a wrapper around that that calls `debug` and adds the log to a centralized log queue.
- Schedule a periodic task that runs every minute.
  - When started, the task extracts all logs that have been made in the last minute from the log queue, and sends them in a batch to the telemetry server.
- Tests to cover the periodic log sending and the logging.

DEN-4505


src/client/specs/telemetry-api.json
```diff
@@ -0,0 +1,18 @@
+{
+  "variables": {
+    "host": "https://telemetry.density.io/v1"
+  },
+  "resources": {
+    "logs": {
+      "batchCreate": {
+        "method": "POST",
+        "url": "{{{host}}}/logs",
+        "headers": {
```
tgross: I don't see it handled explicitly but just in case it hasn't been abstracted away, I wanted to point out this is an authenticated endpoint.

1egoman: Yep, that's abstracted away. The token is set here: https://github.com/DensityCo/dashboard/pull/50/files#diff-2a80aa5e18de6a44b8a2a71e3ed3d55cR38

1egoman:
Oops, need to write tests for this still - that's why CI is failing.

src/client/index.js
```diff
@@ -7,9 +7,10 @@ function errorFormatter(response) {
   });
 }
 
-var core = clientele(Object.assign({}, require('./specs/core-api'), {errorFormatter: errorFormatter}));
-var accounts = clientele(Object.assign({}, require('./specs/accounts-api'), {errorFormatter: errorFormatter}));
-var metrics = clientele(Object.assign({}, require('./specs/metrics-api'), {errorFormatter: errorFormatter}));
+const core = clientele(Object.assign({}, require('./specs/core-api'), {errorFormatter: errorFormatter}));
+const accounts = clientele(Object.assign({}, require('./specs/accounts-api'), {errorFormatter: errorFormatter}));
+const metrics = clientele(Object.assign({}, require('./specs/metrics-api'), {errorFormatter: errorFormatter}));
```
guscost: This API doesn't exist anymore, we should probably remove it.

PR #56, opened 2018-05-02
Ensure that utilization chart never overflows its bounds

1egoman:
Previously, the utilization chart's y axis would always be bounded from `0%` to `100%`. There's a problem with that though if the utilization peaks for a space above `100%`:

screen shot 2018-05-02 at 12 33 01 pm

This PR adds logic to ensure that the largest point on the y axis should either be:
```
1. The largest point on the graph, if larger than 100. (+ 10% in spacing)
2. 100.
```

### DIFF
--- a/src/components/insights-space-detail-utilization-card/index.js
+++ b/src/components/insights-space-detail-utilization-card/index.js
@@ -433,7 +433,14 @@ export default class InsightsSpaceDetailUtilizationCard extends React.Component 
                 formatter: ({value}) => value === 100 ? '100%' : `${value}`,
               })}
               yAxisStart={0}
-              yAxisEnd={100}
+
+              // The largest point on the y axis should either be:
+              // 1. The largest point on the graph, if larger than 100. (+ 10% in spacing)
+              // 2. 100.
+              yAxisEnd={Math.max(
+                Math.max.apply(Math, averageUtilizationDatapointsWithTimestamp.map(i => i.value))+10, /* 1 */
+                100 /* 2 */
+              )}
 
               overlays={[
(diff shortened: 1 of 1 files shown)


src/components/insights-space-detail-utilization-card/index.js
```diff
@@ -433,7 +433,14 @@ export default class InsightsSpaceDetailUtilizationCard extends React.Component
                 formatter: ({value}) => value === 100 ? '100%' : `${value}`,
               })}
               yAxisStart={0}
-              yAxisEnd={100}
+
+              // The largest point on the y axis should either be:
+              // 1. The largest point on the graph, if larger than 100. (+ 10% in spacing)
+              // 2. 100.
+              yAxisEnd={Math.max(
+                Math.max.apply(Math, averageUtilizationDatapointsWithTimestamp.map(i => i.value))+10, /* 1 */
```
bredfield: Why do you use `apply` here vs. just calling the function?

1egoman: `Math.max` accepts a number of arguments that it'll find the max of, ie `Math.max(1, 5, 2, 10)` will return `10`. Therefore, `Math.max(averageUtilizationDatapointsWithTimestamp.map(i => i.value))` won't work, since `Math.max` doesn't accept an array. Therefore, `apply` or argument spreading is required in order to call `Math.max` with an array of values.

bredfield: Cool

guscost: Something something lodash

PR #61, opened 2018-05-09
Visits count adjustment

1egoman:
- Previously, the insights list page was displaying the number of count buckets under `n visits`, not the number of ingresses. This isn't the correct metric in this case.
- Added test to assert behavior.

### DIFF
--- a/src/components/insights-space-list/test.js
+++ b/src/components/insights-space-list/test.js
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { mount, shallow } from 'enzyme';
 import assert from 'assert';
 import sinon from 'sinon';
+import moment from 'moment';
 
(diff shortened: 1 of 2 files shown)


src/components/insights-space-list/test.js
```diff
@@ -670,4 +671,54 @@ describe('insights space list', function() {
...
+          },
+        ],
+        events: {},
+      }}
+      activeModal={{name: null, data: null}}
+    />);
+
+    // Add a couple empty count buckets
+    component.setState({
+      timeSegment: 'WHOLE_DAY',
```
guscost: Should we add a test for workday hours only?

1egoman: Yea good idea

PR #62, opened 2018-05-09
Remove service worker

1egoman:


1egoman:
Actually, not quite done with this. Turns out, this is required too: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#opting-out-of-caching

PR #63, opened 2018-05-10
Sockets reconnect Instability

1egoman:
If the connectionState isn't reset, then attempting to reconnect again
will disconnect the already disconnected connection before reconnection,
which isn't ideal.

### DIFF
--- a/src/helpers/websocket-event-pusher/index.js
+++ b/src/helpers/websocket-event-pusher/index.js
@@ -102,7 +102,7 @@ export default class WebsocketEventPusher extends EventEmitter {
 
         // Clear the interval that sends a ping to the sockets server if it is open.
         if (this.pingIntervalId) {
-          window.clearTimeout(this.pingIntervalId);
+          window.clearInterval(this.pingIntervalId);
         }
 
         if (this.gracefulDisconnect) {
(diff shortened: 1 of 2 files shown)


1egoman:
This is a bugfix, if I remember correctly this was to fix a case where when the socket was already connected and the user ran `.connect()` again, then the connection logic would try to reconnect over and over or something like that.

There is a bit of complexity involved in this process, because there are two different ways of disconnecting from the socket that are required in different cases. The first, a graceful disconnect, occurs when the user logs out (when on the login page, we don't want to have a socket connection open). On a graceful disconnect, the websocket event pusher won't try to reconnect after the `onclose` event has been fired on the websocket instance. The other type of disconnect is a "forceful" disconnect, where the server disconnected from the client but the client should reconnect. In this case, when the `onclose` event is fired, the websocker event pusher will try to reconnect at an exponential backoff.

In order to perform this whole process, there's two bits of state: `connectionState` and `gracefullyDisconnecting`. The first is used to keep track of the current step of the connection process (ie, getting the socket url, connecting to socket, connected to socket) and the second is set so that when `onclose` is fired it knows if it needs to perform the exponential backoff reconnect.

Given this background information (which maybe should be a bit more clear in the code), what part of this system do you not like in particular, and what do you think should be done to address your concerns?

guscost:
Sorry, I missed this comment, and it makes sense. I think I missed the fact that gracefulDisconnect is already a concept and this PR is a modification to its behavior, it's not new.

PR #64, opened 2018-05-30
Live card overflow ellipsis

1egoman:
- Adds ellipsis to live space card, which more gracefully handles long space names.

screen shot 2018-05-30 at 3 37 46 pm

### DIFF
--- a/src/components/live-space-card/index.js
+++ b/src/components/live-space-card/index.js
@@ -29,16 +29,7 @@ export default function SpaceCard({
     const capacityPercent = space.capacity ? (space.currentCount / space.capacity) * 100 : null;
     return <Card className="space-card">
       <CardHeader className="space-card-header">
-        <span className="space-card-header-name">
-          {space.name}
-          <span
-            className="space-card-header-name-link"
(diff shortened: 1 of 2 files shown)


src/components/live-space-card/index.js
```diff
@@ -29,16 +29,7 @@ export default function SpaceCard({
     const capacityPercent = space.capacity ? (space.currentCount / space.capacity) * 100 : null;
     return <Card className="space-card">
       <CardHeader className="space-card-header">
-        <span className="space-card-header-name">
-          {space.name}
-          <span
-            className="space-card-header-name-link"
-            onClick={() => { window.location.href = `#/spaces/insights/${space.id}`; }}
```
guscost: We're removing this link?

1egoman: Yea, I mentioned it to kyle and given that it doesn't work all that particularly well on mobile (activates on hover), it makes the text longer (and harder to ellipsize), and we haven't heard anything positive about it I removed it. If you feel differently though, lmk.

guscost: Not particularly. It would be nice to have a way to jump to a space details page from this page but we should think it through.

PR #73, opened 2018-07-02
Dashboard changes to support demo account

1egoman:
- Adjusted error handler - was previously redirecting to the login page when any 403 happened, now redirects when a response of `{"detail":"Incorrect authentication credentials"}` is received. Without this, anytime the demo user does anything they aren't supposed to, they get redirected to the login page.

- Disable the onboarding navbar item when the logged in user is a demo user.

<br/>

screen shot 2018-07-02 at 8 37 35 am

- Remove the `Edit` and `Change Password` links on the account page when the logged in user is a demo user.

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -7,19 +7,25 @@ let store;
 function setStore(s) { store = s; }
 
 async function errorHandler(response) {
-  const data = await response.json();
+  try {
+    const data = await response.json();
 
-  // If the user received a 403 with a body of 'invalid authentication credentials' in response to
-  // any request, send them to the login page.  Redirect the user to the login page and remove the
-  // bad session token from the reducer.
-  if (response.status === 403 && data.detail === 'Incorrect authentication credentials.') {
-    store.dispatch(userError(`Login session has expired or is invalid. Please login again.`));
-    store.dispatch(sessionTokenUnSet());
-    window.location.href = '#/login';
+    // If the user received a 403 with a body of 'invalid authentication credentials' in response to
+    // any request, send them to the login page.  Redirect the user to the login page and remove the
+    // bad session token from the reducer.
+    if (response.status === 403 && data.detail === 'Incorrect authentication credentials.') {
+      store.dispatch(userError(`Login session has expired or is invalid. Please login again.`));
+      store.dispatch(sessionTokenUnSet());
+      window.location.href = '#/login';
+    }
+
+    // If the response wasn't a 403, then return the best representation of the error.
+    return data.detail || data.error || JSON.stringify(data);
+  } catch (err) {
+    // Handle requests that don't contain json.
+    const text = await response.text();
+    return text;
   }
-
-  // If the response wasn't a 403, then return the best representation of the error.
-  return data.detail || data.error || JSON.stringify(data);
 }
 
 const core = clientele({...require('./specs/core-api'), errorFormatter: errorHandler});
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -7,17 +7,18 @@ let store;
 function setStore(s) { store = s; }
 
 async function errorHandler(response) {
-  // If the user received a 403 in response to any request, send them to the login page.
-  // Redirect the user to the login page and remove the bad session token from
-  // the reducer.
-  if (response.status === 403) {
+  const data = await response.json();
+
+  // If the user received a 403 with a body of 'invalid authentication credentials' in response to
+  // any request, send them to the login page.  Redirect the user to the login page and remove the
+  // bad session token from the reducer.
+  if (response.status === 403 && data.detail === 'Incorrect authentication credentials.') {
     store.dispatch(userError(`Login session has expired or is invalid. Please login again.`));
     store.dispatch(sessionTokenUnSet());
     window.location.href = '#/login';
   }
 
   // If the response wasn't a 403, then return the best representation of the error.
-  const data = await response.json();
   return data.detail || data.error || JSON.stringify(data);
 }
 
--- a/src/components/account/index.js
+++ b/src/components/account/index.js
@@ -31,17 +31,17 @@ export class Account extends React.Component {
       passwordConfirmation: '',
 
       // Initialize with a prop passing the initial value from the store
-      fullName: this.props.user ? this.props.user.fullName : '',
-      nickname: this.props.user ? this.props.user.nickname : '',
-      email: this.props.user ? this.props.user.email : '',
+      fullName: this.props.user.data ? this.props.user.data.fullName : '',
+      nickname: this.props.user.data ? this.props.user.data.nickname : '',
+      email: this.props.user.data ? this.props.user.data.email : '',
     };
   }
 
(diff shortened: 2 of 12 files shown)


src/client/index.js
```diff
@@ -7,17 +7,18 @@ let store;
...
-  // If the user received a 403 in response to any request, send them to the login page.
-  // Redirect the user to the login page and remove the bad session token from
-  // the reducer.
-  if (response.status === 403) {
+  const data = await response.json();
+
+  // If the user received a 403 with a body of 'invalid authentication credentials' in response to
+  // any request, send them to the login page.  Redirect the user to the login page and remove the
+  // bad session token from the reducer.
+  if (response.status === 403 && data.detail === 'Incorrect authentication credentials.') {
```
guscost: Why check for both the text and the code here?

src/client/index.js
```diff
@@ -7,17 +7,18 @@ let store;
 function setStore(s) { store = s; }
 
 async function errorHandler(response) {
-  // If the user received a 403 in response to any request, send them to the login page.
-  // Redirect the user to the login page and remove the bad session token from
-  // the reducer.
-  if (response.status === 403) {
+  const data = await response.json();
```
guscost: This, and the old behavior as well, could fail if the server returns a 500 or something else w/ no valid json (nginx 410, etc).

src/components/dev-token-update-modal/index.js
```diff
@@ -98,7 +99,7 @@ export default class TokenUpdateModal extends React.Component {
 
             <p>
               The act of removing a token is irreversible - ie, you might have the token built into
-              a compiled thing somewhere that would be hard / impossible to reassign. Type in the
+              a piece of software that might be difficult to reassign. Type in the
```
guscost: how about "difficult to update"?

src/components/dev-webhook-list/index.js
```diff
@@ -83,7 +83,7 @@ export function WebhookList({
           <h1 className="webhook-list-header-text">Webhooks</h1>
           <DescriptionModal>
             <p>
-              Webhooks allow us to push you data whenever your Density sensors count an event. Create a webhook, tell us what endpoint to hit, and we'll send you a payload over HTTP.
+              Webhooks allow us to push your data whenever your Density sensors count an event. Create a webhook, tell us what endpoint to hit, and we'll send you a payload over HTTP.
```
guscost: Over HTTP(S)? Not sure if we should suggest HTTP although it's not necessarily significant in context here. Maybe "we'll send you an HTTP request for each event"

src/components/insights-space-detail-utilization-card/index.js
```diff
@@ -79,8 +79,8 @@ export default class InsightsSpaceDetailUtilizationCard extends React.Component
       data: null,
       dataSpaceId: null,
 
-      startDate: moment.utc().tz(this.props.space.timeZone).subtract(7, 'days').startOf('day').format(),
-      endDate: moment.utc().tz(this.props.space.timeZone).subtract(1, 'day').endOf('day').format(),
+      startDate: moment.utc().tz(props.space.timeZone).subtract(1, 'week').startOf('week').format(),
```
guscost: I don't understand the reasoning for this and the related timestamp changes (are these really from the demo user PR?) but nothing seems obviously wrong here.

src/components/insights-space-list/index.js
```diff
@@ -106,7 +106,7 @@ export class InsightsSpaceList extends React.Component {
 
       // NOTE: The below times don't have timezones. This is purposeful - the
       // `core.spaces.allCounts` call below can accept timezoneless timestamps, which in this case,
-      // is desired so that we can get from `startTime` to `endTime` in the timezone of each space,
+      // is desired so that we can get from `startDate` to `endDate` in the timezone of each space,
```
guscost: The convention in the API/docs is still `start_time` and `end_time`, not sure if it matters here but why the change?

src/components/insights-space-list/index.js
```diff
@@ -119,19 +119,20 @@ export class InsightsSpaceList extends React.Component {
       //  9  10
       //
       // ie, if the current date is the 10th, then the last full week goes from the 3rd to the 8th.
-      let startTime,
-          endTime = moment.utc().subtract(1, 'week').endOf('week').subtract(1, 'day').format('YYYY-MM-DDTHH:mm:ss');
+      let startDate,
+          endDate = moment.utc().subtract(1, 'week').endOf('week').endOf('day').format('YYYY-MM-DDTHH:mm:ss');
+
       if (this.state.dataDuration === DATA_DURATION_WEEK) {
-        startTime = moment.utc().subtract(1, 'week').startOf('week').subtract(1, 'day').startOf('day').format('YYYY-MM-DDTHH:mm:ss');
```
guscost: These don't have a local TZ, is that intentional? Perhaps it's missing?

src/reducers/user/index.js
```diff
@@ -19,6 +20,8 @@ export function user(state=initialState, action) {
     return {...state, loading: false, data: {...state.data, ...objectSnakeToCamel(action.item)}};
   case USER_ERROR:
     return {...state, loading: false, error: action.error};
+  case SESSION_TOKEN_UNSET:
```
guscost: Nice 👍

src/components/account/index.js
```diff
@@ -77,7 +77,7 @@ export class Account extends React.Component {
           {this.state.mode === EDIT ? 'Edit Account' : 'Account'}
 
           {/* Edit / Cancel button */}
-          <ModalHeaderActionButton
+          {user && !user.isDemo ? <ModalHeaderActionButton
```
guscost: Small nitpick: Elsewhere in the app this prop is `connect`ed as `state.user`, but here it's `state.user.data`. It obviously works fine though.

src/client/index.js
guscost: Nevermind! I understand based on the description. Perhaps we should be using 401 here but I don't think that's a DRF convention.

src/components/insights-space-list/index.js
1egoman: Yep, it intentional - it's the `spaces/counts/` endpoint! Documented above: https://github.com/DensityCo/dashboard/blob/demo-data-fixes/src/components/insights-space-list/index.js#L107

src/components/insights-space-list/index.js
1egoman: I wanted to make the language in the utilization card and the insights space list consistent, and one was using `{start,end}Time` and one was using `{start,end}Date`. I picked the date ones because in this context the date is whats being used in the calculations (ie, the calculations never get below the granularity of a day). Maybe they should be `{start,end}Timestamp` or something like that?

guscost: It's no big deal right now, and this makes sense. We might want to add recommendations for this situation to a brand style guide if/when we end up making one.

src/components/insights-space-list/index.js
guscost: 👍

PR #79, opened 2018-09-27
Sean hotfix

1egoman:
This contains a few small changes I made before and during meeting with @sangelone.


sangelone:
Let's show these to users/French, especially the start/end in space's timezone and moving back the default range a day.

PR #82, opened 2018-10-02
Only draw data that fits within the foot traffic chart

1egoman:
This was an issue because it turns out that if data is requested for a time period that doesn't subdivide into even intervals, then more data than expected will be returned. Therefore, the chart should filter the data prior to rendering it to only display data that fits within its bounds. Furthermore, if removing a final section of data would leave a gap, fill in that gap by adding a final point at the end of the chart on days that aren't the current day (the current day hasn't finished yet!)

Old behavior:
screen shot 2018-10-02 at 3 55 14 pm

New behavior:
screen shot 2018-10-02 at 3 56 02 pm

### DIFF
--- a/src/components/insights-space-detail-foot-traffic-card/index.js
+++ b/src/components/insights-space-detail-foot-traffic-card/index.js
@@ -92,6 +92,7 @@ export default class InsightsSpaceDetailFootTrafficCard extends React.Component 
         date,
         timeSegment,
         timeSegmentGroup,
+        dataSpaceId: space.id,
       }, () => this.fetchData());
     }
   }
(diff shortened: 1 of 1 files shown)


src/components/insights-space-detail-foot-traffic-card/index.js
```diff
@@ -134,6 +121,47 @@ export default class InsightsSpaceDetailFootTrafficCard extends React.Component
...
+      // In addition, don't attempt to "fill in" data for the current day, as it hasn't happened
+      // yet!
+      if (chartData.length > 0 && !isToday) {
+        chartData.push({
+          timestamp: endTime,
+          value: chartData[chartData.length-1].value,
+        });
+      }
+
+      min = Math.min.apply(Math, data.map(i => i.count));
```
sangelone: Will this conflict with changes in #81?

1egoman: Good point, they might. We'll see when they are merged.

PR #88, opened 2018-10-10
Error helper den 6542

1egoman:
Many of the action creators in the dashboard contain code that looks like this:
```javascript
// Load a list of all spaces
try {
  const spaces = await core.spaces.list(); 
  dispatch(collectionSpacesSet(spaces.results));
} catch (err) {
  dispatch(collectionSpacesError(`Error loading space: ${err}`));
}
```

There's a pretty big bug with this though - the `dispatch` call is within the try block. Since the dispatch eventually ties into react through react-redux, any errors that are thrown in the component that are not handled in another way are not shown in the console and instead processed and stored into the reducer. This is not a good.

This is a work in progress PR to instead use a specially build "error helper" which contains an "else" branch that will fire when no error was thrown in the "try" branch, similarly to the way it works in python:
```javascript
return errorHelper({
  try: () => core.webhooks.list(),
  catch: error => {
    dispatch(collectionWebhooksError(error));
  },
  else: async request => {
    const webhooks = await request;
    dispatch(collectionWebhooksSet(webhooks.results));
  },
});
```

### DIFF
--- a/src/helpers/error-helper/index.js
+++ b/src/helpers/error-helper/index.js
@@ -1,4 +1,4 @@
-export default function errorHelper(params /* {try, catch, finally, else} */) {
+export default async function errorHelper(params /* {try, catch, finally, else} */) {
   let result = undefined,
       errorThrown = false;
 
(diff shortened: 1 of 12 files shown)


src/helpers/error-helper/index.js
```diff
@@ -0,0 +1,30 @@
+export default async function errorHelper(params /* {try, catch, finally, else} */) {
```
guscost: Could we add some comments in/above this function explaining why it's a necessary abstraction? I know we talked about this but can't easily understand why we need both `finally` and `else` (what I assume is the reason we can't easily use only the built-in control flow blocks).

PR #93, opened 2018-11-12
Remove lab from environment switcher

1egoman:


sangelone:
And now it looks good too! 👍

PR #94, opened 2018-11-13
Add hourly breakdown report and calculations

1egoman:
What's left:
- [x] Review of logic for `Hourly Breakdown` in https://confluence.density.io/display/WMA/Report+Definitions+v1
- [x] PR code review, and make sure the logic matches the proposal in that document
- [x] Add `HOURLY_BREAKDOWN` report type to core api


src/actions/collection/dashboards/report-calculations/hourly-breakdown.js
```diff
@@ -0,0 +1,88 @@
...
+      id: report.settings.spaceId,
+      interval: '1h',
+      start_time: formatInISOTimeAtSpace(timeRange.start, space),
+      end_time: formatInISOTimeAtSpace(timeRange.end, space),
+      page,
+      page_size: 1000,
+    });
+  });
+
+  // Filter data to remove all buckets with a timerange outside of the start and end hours speicfied
```
guscost: "with a timestamp..."?

src/actions/collection/dashboards/report-calculations/hourly-breakdown.js
```diff
@@ -0,0 +1,88 @@
...
+  // in the report settings
+  data = data.filter(bucket => {
+    const timestamp = parseISOTimeAtSpace(bucket.timestamp, space);
+    return report.settings.hourStart <= timestamp.hours() && timestamp.hours() <= report.settings.hourEnd;
```
guscost: Possibly should not be inclusive on the `hourEnd` comparison?

src/actions/collection/dashboards/report-calculations/hourly-breakdown.js
```diff
@@ -0,0 +1,88 @@
...
+      return !isWeekend;
+    });
+  }
+
+  // Finally, group together all count buckets fetched into groups, one per day.
+  const bucketsByDay = data.reduce((acc, bucket) => {
+    const date = parseISOTimeAtSpace(bucket.timestamp, space).format('YYYY-MM-DD');
+    const dateIndex = acc.findIndex(d => d.date === date);
+    if (dateIndex >= 0) {
+      return [
```
guscost: There is no way this is the simplest way to do this (or we should use RxJS/lodash if so). It doesn't look incorrect but it does look like one of the purposely-confusing math problems that you do for extra credit.

src/actions/collection/dashboards/report-calculations/hourly-breakdown.js
1egoman: Not sure on this, @bredfield can you comment?

1egoman:
```
ben [3:32 PM]
I believe each hour should be, for example, 12:00 -> 12:59:59

So non-inclusive
ryan [3:32 PM]
Non inclusive on both sides, or just at the end?
ben [3:33 PM]
Start should be inclusive
If an event happens at 12:00:00.000 it should be included in the 12 -> "13" bucket
ryan [3:33 PM]
Ok sounds good
```

src/actions/collection/dashboards/report-calculations/hourly-breakdown.js
1egoman: Chatted in real life, adding a bit of documentation but not a ton of concern over this one

PR #98, opened 2018-11-14
code: fix peak times in comparison chart

guscost:


1egoman:
@guscost Be sure to update https://confluence.density.io/pages/viewpage.action?spaceKey=WMA&title=Report+Definitions+v1 prior to merging!

guscost:
Left a comment, but the definition is pretty vague. I don't agree with this "fixed" method either, but am happy to copy out that comment into the definition somewhere.

1egoman:
Ok, can you update the definition once you've come to something that you're satisfied with? I'm trying to minimize as much as possible drift between that document and the code that goes along with it.

guscost:
Yes, I think leaving it as a comment for now is OK, since this document (and our analytical methods for these reports) may still be evolving for some time.

guscost:
OK, updating the description instead.

PR #99, opened 2018-11-15
Updated Live Screen to be max-height 300px for small screens

beweinreich:
This allows the top part to stay visible until the screen shrinks to 300px in height. This was a request from Pier when demonstrating the live screen to a customer with a smaller screen. It kept hiding the top part and he was confused why it wouldn't show. 

I'm comfortable with this look.

screen shot 2018-11-14 at 5 48 15 pm


1egoman:
Though @beweinreich as always, you need to work on your commit messages

PR #100, opened 2018-11-15
Render wasted space report and do its calculations

1egoman:
1. I'm not sure how useful this report is currently. In general, the notion of a percentage being associated with utilization seems like it doesn't answer the full question that this report was built to answer. Somehow, it seems we may want to elaborate on this "x% of space is underutilized" - maybe list out the average utilization percentages of the spaces? Idk, something to explain why we see this space as underutilized.
2. I left a few comments in the code that I will call out in a review that I don't particularly like about how this data fetching code was written. They are worth looking at.


src/actions/collection/dashboards/report-calculations/wasted-space.js
```diff
@@ -0,0 +1,83 @@
...
+  const spaces = allSpaces.filter(space => report.settings.spaceIds.indexOf(space.id) >= 0);
+
+  if (spaces.length === 0) {
+    throw new Error(`The given space ids weren't found.`);
+  }
+
+  // Fetch all count data within the time range, specifying an interval of 1 hour
+  const data = {};
+  await Promise.all(spaces.map(async space => {
+    const timeRange = convertTimeRangeToDaysAgo(space, report.settings.timeRange);
```
1egoman: I would have liked to use the `/spaces/counts` endpoint here, but that's not going to work. With our recent focus on daylight savings, the concept behind this endpoint in the usecase used most often (passing timezone-less timestamps to the start and end query params, so that all counts in the local times of each space would be returned) is a bit flawed - this is because a daylight savings boundary could occur in the middle of one space's time zone, and not in the other, and return a different set of data.

So, to get around this, I used the `/spaces/:id/counts` endpoint called multiple times. This is not a great idea though, because a call will be made per space, and potentially if a lot of spaces are used, overwhelm the gunicorn workers. This happened on the insights page way back when and was the reason we made the `/spaces/counts/` endpoint in the first place.

src/actions/collection/dashboards/report-calculations/wasted-space.js
```diff
@@ -0,0 +1,83 @@
...
+  const overutilizedPercent = (
+    spaceAverageUtilization.filter(i => normalOverutilizedThreshold < i).length
+      / spaceAverageUtilization.length
+  ) * 100;
+
+  // XXX
+  // How do we determine which time range to display in the header of the report, given that the
+  // time range to be rendered requires a space to be provided?
+  // Answer: at the moment, display the time range for the first space.
+  const displayedTimeRange = convertTimeRangeToDaysAgo(spaces[0], report.settings.timeRange);
```
1egoman: This probably isn't the correct way to get the time range to render in the header of the report card. The problem is that in order to convert `NEXT_WEEK` or `LAST_7_DAYS` into a date range, we need a time zone, and multiple spaces that are queried could have differing time zones, causing issues.

src/actions/collection/dashboards/report-calculations/wasted-space.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  // Fetch all count data within the time range, specifying an interval of 1 hour
+  const data = {};
+  await Promise.all(spaces.map(async space => {
+    const timeRange = convertTimeRangeToDaysAgo(space, report.settings.timeRange);
+    data[space.id] = await fetchAllPages(page => {
+      return core.spaces.counts({
+        id: space.id,
+        interval: '1h',
```
1egoman: Does sampling the data at an hour vs at a differing interval cause the average to be skewed? This is something that should be confirmed.

1egoman: Ok, that idea seems to work, I'm going to go ahead with it

src/actions/collection/dashboards/report-calculations/wasted-space.js
guscost: This will be OK when the calls are faster, and I can't think of an alternative that is both correct and manageable. If we need it today we crank those EC2+RDS knobs, maybe.

src/actions/collection/dashboards/report-calculations/wasted-space.js
1egoman: On second thought, this idea is bad, because the utilization returned is for the start of the bucket, not across the whole bucket.

1egoman:
@bredfield Migrated to using 5m buckets - anything other review items?

PR #101, opened 2018-11-15
Feat: more dst fixing

guscost:


src/helpers/space-time-utilities/index.js
```diff
@@ -160,11 +160,11 @@ export function splitTimeRangeIntoSubrangesWithSameOffset(space, start, end, par
       gap: false
     });
 
-    // Adjust the next interval to shift to align with the new offset, if requesting weeks or days
-    if (['w', 'd'].includes(params.interval.slice(-1))) {
+    // If querying days or weeks, shift the next interval to align with the new offset
+    if (['d', 'w'].includes(params.interval.slice(-1))) {
```
1egoman: FYI, `Array.prototype.includes` isn't supported in ie 11, so we probably need a polyfill if that's something you want to use

guscost: ah, i'll use indexof

PR #103, opened 2018-11-26
code: add total visits rollup report

guscost:
## Example Settings

```
{
  "mode": "MOST_VISITED",
  "space_ids": [
    "spc_578609895201309436",
    "spc_578610433024328307",
    "spc_578610047240634650",
    "spc_578610642626281829",
    "spc_578609560999166590"
  ],
  "time_range": "LAST_WEEK",
  "time_segment_group_id": "tsg_581960155378221081"
}
```

## Result

screen shot 2018-11-26 at 2 09 45 pm


bredfield:
Minor request: If no pixels are rendered in the bar chart, remove the margin left from the count label

src/actions/collection/dashboards/report-calculations/total-visits-rollup.js
```diff
@@ -0,0 +1,90 @@
...
+    /* 'spc_xxx': [{timestamp: "...", interval: {...}, count: 100}], */
+  };
+  await Promise.all(spaces.map(async (space, index) => {
+    const timeRange = convertTimeRangeToDaysAgo(space, report.settings.timeRange);
+    const responseData = await fetchAllPages(page => {
+      return core.spaces.counts({
+        id: space.id,
+        // Calculate an `interval` parameter that guarantees a single count bucket be returned
+        // between the start and end times.
+        interval: `${Math.ceil(
```
1egoman: I think getting the single bucket is ok in this case because we're only using it for `entrances`, and not a value that depends on the start time of the bucket?

guscost: Believe so, as long as the range includes what we need for this report, it should return the correct total of all relevant time segments.

PR #105, opened 2018-11-26
Add logic to render last point after the last bucket

1egoman:
When rendering all points in the graph in the average time segment breakdown report, points are rendered at the start of each count bucket. However, when done this way, there's a bit of a gap at the end of the graph because the final point is at the start of the final bucket of data, not at the end. Therefore, fetch the next count bucket after the final bucket and use this value to figure out the "final" count to plot on the graph.


src/actions/collection/dashboards/report-calculations/average-time-segment-breakdown.js
```diff
@@ -155,7 +156,34 @@ export default async function averageTimeSegmentBreakdown(report) {
...
+    id: report.settings.spaceId,
+    interval: '5m',
+    order: 'asc',
+
+    start_time: formatInISOTimeAtSpace(timeRange.end, space),
+    end_time: formatInISOTimeAtSpace(
+      timeRange.end.clone().add(5, 'minutes'),
+      space,
+    ),
+    time_segment_groups: report.settings.timeSegmentGroupId,
```
guscost: What if the time segment doesn't include this "final" bucket? It seems like a more correct approach would be to use the "last count" for the last bucket, which the API does use internally but is not part of the count endpoint response.

guscost: And I realize I'm requesting changes to the Core API too, perhaps @beweinreich can weigh in?

1egoman: Yea that's a good point 😦

bredfield: @guscost can you propose an API adjustment that wouldn't complicate the response body?

guscost: it could be just a new `final_count` key either in the default response or w/ a flag (this example also adds an `initial_count` key which is redundant, but so is `timestamp`/`start`):

```
{
      "timestamp": "2017-09-17T08:00:00.000Z",
      "count": 0,
      "interval": {
        "start": "2017-09-17T08:00:00.000Z",
        "end": "2017-09-18T07:59:59.999Z",
        "initial_count": 0,
        "final_count": 0,
        "analytics": {
          "min": 0,
          "max": 43,
          "events": 468,
          "entrances": 234,
          "exits": 234
        }
      }
    }
```

PR #108, opened 2018-11-27
React 16 take two

1egoman:
**NOTE**: there are some styling issues that came up when I upgraded the ui card component. I didn't fix these right now as I wasn't sure if the work wouldn't be worthwhile with the upcoming work rob is doing.

Perform a number of refactoring items on the dashboard.
- Upgrade to react 16.6.3 (latest as of the middle of last week). This gives us a lot of new features, such as fragments, error boundaries, hooks, and much, much more.
- Upgrade to create-react-app 2.1.1
- Migrate all explore pages to use a new reducer called the `exploreData` reducer.
  - Create the reducer
  - Migrate each explore module to use this reducer
    - Space list
    - Daily Metrics card
    - Utilization card
    - Foot traffic card
    - Daily raw events card
- Remove the environment page, as it's not used by external parties
  - Removes multiple thousands of lines of code!
- Upgrade enzyme, and adjust all tests to use the new enzyme version.


package.json
```diff
@@ -87,5 +86,11 @@
     "make-component": "./utilities/make-component",
     "make-helper": "./utilities/make-helper",
     "make-reducer": "./utilities/make-reducer"
-  }
+  },
+  "browserslist": [
```
guscost: What's this?

src/actions/route-transition/explore-space-trends.js
```diff
@@ -5,32 +5,251 @@ import collectionSpacesError from '../collection/spaces/error';
...
-    } catch (err) {
-      dispatch(collectionTimeSegmentGroupsError(`Error loading time segments: ${err.message}`));
+      data = await requestCountsForLocalRange(
+        space,
+        startTime,
+        endTime,
+        {
+          interval: '1d',
+          order: 'asc',
+          page_size: 1000,
```
guscost: Doesn't matter here but we can request 5000 now

src/components/account-setup-doorway-detail/_styles.scss
```diff
@@ -57,7 +57,7 @@
   margin-bottom: 20px;
   color: $gray-dark;
   font-size: $font-size-base;
-  font-weight: lighter;
+  font-weight: 400;
```
guscost: What does this have to do with anything?

src/actions/route-transition/explore-space-trends.js
1egoman: Ah yea, I think I botched the merge on this, will move them all back to 5000

guscost:
screen shot 2018-11-30 at 3 48 43 pm

Is this supposed to look different?

package.json
1egoman: Was part of the migration instructions, I think it relates to which browsers it targets?

src/components/account-setup-doorway-detail/_styles.scss
1egoman: Some browsers render `lighter` with a slimmer font weight, it turns out. /shrug

1egoman:
From PR description:
```
NOTE: there are some styling issues that came up when I upgraded the ui card component. I didn't fix these right now as I wasn't sure if the work wouldn't be worthwhile with the upcoming work rob is doing.
```

guscost:
screen shot 2018-11-30 at 3 50 56 pm

horizontal scroll bar is back

guscost:
screen shot 2018-11-30 at 3 51 56 pm

raw events list flickers and displays "0 events" in between loading pages

1egoman:
Yep more of that same comment I posted above, I'm not sure what the priories are and if we should fix these things now or if we're going to update the ui first.

guscost:
does the card markup really need to be upgraded at all? seems like this update has a dependency on other work that we haven't even started.

guscost:
OK reverting the card version to 1.1.0 fixed those issues. There's still a flicker when loading raw events but I don't think that's a big deal. I think we can ship and deal with the card thing separately.

src/components/account-setup-doorway-detail/_styles.scss
guscost: 🤕

PR #118, opened 2018-12-12
Dashboard expanded report

1egoman:
This pull request implements the other half to https://github.com/DensityCo/ui/pull/78, which permits reports to expand.


src/components/dashboard/_styles.scss
```diff
@@ -157,3 +166,84 @@
   font-size: 24px;
   color: $gray-cinder;
 }
+
+
+@keyframes modalAnimateIn {
```
1egoman: These animations may need some help, I'm not all that great at getting these pixel perfect

src/components/report/index.js
```diff
@@ -0,0 +1,191 @@
...
+// There are three concepts that relate to reports:
+// - The "report", which is the value returned from the api when we make a request to GET /reports
+// or GET /reports/:id
+// - The "report data", which is generated by a "report calculation" for the given report type. This
+// data contains the values that are used to render the graphs and visualizations within the report.
+// - The "report settings", which contains a way to change props passed to a report given parameters
+// configured at the time of report render. A great example - two reports can render the same report
+// data, but one may be "expanded", and one may not be. This shouldn't be in the "report data",
+// since it can change depending on situation the report is rendered in.
+
```
1egoman: Inline summary repeated here, worth reading:
```
// A Report is a small card shaped element rendered to visualize density data.
// There are three concepts that relate to reports:
// - The "report", which is the value returned from the api when we make a request to GET /reports
// or GET /reports/:id
// - The "report data", which is generated by a "report calculation" for the given report type. This
// data contains the values that are used to render the graphs and visualizations within the report.
// - The "report settings", which contains a way to change props passed to a report given parameters
// configured at the time of report render. A great example - two reports can render the same report
// data, but one may be "expanded", and one may not be. This shouldn't be in the "report data",
// since it can change depending on situation the report is rendered in.
```

src/components/report/index.js
```diff
@@ -0,0 +1,191 @@
...
+// Add new report types here, mapping to their component.
+// NOTE: if the report is expandable, add the correct properties to the REPORT_TYPE_TO_SETTINGS
+// structure below to!
+const REPORT_TYPE_TO_COMPONENT = {
```
1egoman: `REPORT_TYPE_TO_COMPONENT` has moved into this `Report` component as now it handles rendering reports in all contexts.

src/components/report/index.js
```diff
@@ -0,0 +1,191 @@
...
+
+function Report({
+  report,
+  reportData,
+  expanded,
+
+  dashboardReportGridIdentityValue,
+  onOpenReportExpandedModal,
+}) {
+  const REPORT_TYPE_TO_SETTINGS = {
```
1egoman: There's this new `REPORT_TYPE_TO_SETTINGS` concept which defines props that are passed to a report componentt when it's rendered in different modes.

First, is the name `REPORT_TYPE_TO_SETTINGS` make sense? <aybe it should be something like `REPORT_TYPE_TO_VISUAL_DISPLAY_INFORMATION` or something like that?

Second, this now introduces a new place where props to a report can come from, which means that  whatever data is returned out of a report calculation is no longer the only thing going to a report. I'm on the fence on this idea - I really like the simplicity of the original solution, but I'm not sure how to extend it to set dynamic report props without re-performing data calculations when the modal opens.

src/components/report/index.js
```diff
@@ -0,0 +1,191 @@
...
+    return (
+      <ReportComponent
+        key={report.id}
+        title={report.name}
+
+        /* inject return data that calculated by data calculations as props */
+        {...reportData.data}
+
+        /* inject custom props when a report is expandable */
+        {...(() => {
```
1egoman: This is where the data in the `REPORT_TYPE_TO_SETTINGS` structure actually makes its way into the props of the report component. Note that any props injected in this way override the report calculation props.

src/components/dashboard/_styles.scss
guscost: Will QA this and address if necessary in QA.

src/components/report/index.js
guscost: Could it be a single additional object prop like `context` which is reserved for this use case? Seems like that wouldn't be that much of a problem.

1egoman: I like this idea. Once concern with this though is that we may have to change the signature of a report's props if we want to change a if a value is in the "context"

guscost: I don't understand, you mean changing the schema/PropTypes including the nested context values?

1egoman: Example: lets say that that `propA` at the moment should be `50` when expanded, and `20` when not. It would in this paradigm then live inside of `context`, so the context prop would look like `{propA: 50}` or `{propA: 20}`. Now, if I wanted to update the logic so that `propA` was always `20`, it would require two steps: moving `propA` out of `context`, and this would mean that the report would need a major version bump as the format of its props would not be backwards compatible.

This isn't a problem, but just something we should be aware of if we're going to go down this route.

guscost: ok yes, that is a wrinkle, but doesn't seem like the wrong thing to do in that case.

PR #121, opened 2018-12-19
Navbar Dropdown

1egoman:
screen shot 2018-12-19 at 10 56 55 am

For more, see the preview link.


guscost:
the click behavior is fixed, but it turns out safari doesn't work as expected in general with keyboard navigation (you need to do option+tab without a change to preferences)

PR #123, opened 2018-12-29
TypeScript

guscost:


src/components/account-setup-doorway-detail/index.tsx
```diff
@@ -41,8 +41,8 @@ export class AccountSetupDoorwayDetail extends React.Component {
 
       // These are string representations of the width and height. Before they leave the component,
       // these values are converted to centimeters and added back into `state.doorway.{width,height}`.
-      inputWidth: this.props.initialDoorway.environment && this.props.initialDoorway.environment.width ? window.parseFloat(this.props.initialDoorway.environment.width, 10) * CENTIMETERS_PER_METER : '',
-      inputHeight: this.props.initialDoorway.environment && this.props.initialDoorway.environment.height ? window.parseFloat(this.props.initialDoorway.environment.height, 10) * CENTIMETERS_PER_METER : '',
+      inputWidth: this.props.initialDoorway.environment && this.props.initialDoorway.environment.width ? parseFloat(this.props.initialDoorway.environment.width) * CENTIMETERS_PER_METER : '',
```
1egoman: `parseFloat`/`parseInt` (there's a bunch in this file) needs to have the `radix` 2nd param, there's some edge cases if it's not included: https://coderwall.com/p/4eaixa/parseint-can-be-dangerous

guscost: From Slack: `parseFloat` doesn't actually use a radix argument, so typescript expects one param for it.

PR #124, opened 2019-01-02
Navbar dropdown (alternate event handlers)

guscost:


guscost:
ah this has other problems w/ tab indexing

PR #128, opened 2019-01-08
Sensors page

beweinreich:
No frills, no fuss.

This is a list of your sensors, sorted by which are down (error state). It's hidden for now until we get some feedback.

screen shot 2019-01-08 at 3 28 41 pm


beweinreich:
eh @1egoman hold off on reviewing this. gonna make the background white

1egoman:
@beweinreich `we dont need a test for sensors page since its not official` I don't really agree with this... Tests should be there to help you ensure that functionality works long term, and knowing how these things usually go, most likely not doing these now will mean that they will never get done.

PR #129, opened 2019-01-10
New comparison report

guscost:


src/actions/collection/dashboards/report-calculations/comparative-performance.ts
```diff
@@ -40,100 +36,98 @@ export default async function comparativePerformance(report) {
...
+  const lastStartDate = getCurrentLocalTimeAtSpace(space).startOf(unit).subtract(1, unit);
+
+  // Build array of periods
+  const mode = TYPE_TO_MODE[report.settings.type];
+  const size = report.settings.size || (mode === COMPARATIVE_WEEK ? 4 : 3);
+  const periods: any[] = [];
+  for (let i = 0; i < size; i++) {
+    const startDate = lastStartDate.clone().subtract(i, unit);
+    const endDate = startDate.clone().endOf(unit);
+    periods.push({ startDate, endDate });
```
1egoman: Nitpick and not something I feel strongly about - maybe use `unshift` here instead of push so you don't have to reverse it below?

guscost: Nice, sounds good.

PR #132, opened 2019-01-17
Add error boundary around each report

1egoman:
If a report throws an error while rendering, then render a "report error" state instead of letting the error make its way to react and causing the whole app to unmount. This gets rid of the issue where a report failing to render would cause the whole application to unmount and a white screen shown instead.

### DIFF
--- a/src/components/report/index.tsx
+++ b/src/components/report/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { connect } from 'react-redux';
 
 import REPORTS, { DEFAULT_REPORT_SETTINGS } from '../../reports';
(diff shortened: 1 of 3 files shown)


src/components/report/index.tsx
```diff
@@ -21,85 +21,124 @@ import showModal from '../../actions/modal/show';
...
+      break;
+    }
+
     return (
-      <ReportComponent
-        key={report.id}
-        title={report.name}
-        {...reportProps}
-      />
+      <span>Unknown report state {reportData.state}</span>
```
guscost: I'd put this in the default case, but I guess this is technically safer.

guscost:
Needs to be moved into https://github.com/DensityCo/reports

PR #135, opened 2019-01-29
Convert reports to use @density/reports

1egoman:
1. Remove all report calculations and dependencies on reports
2. Add in `@density/reports`. Update all references of `REPORTS` and `Report` to point to this package.
3. Add a few types in order to appease typescript


src/setupTests.ts
```diff
@@ -1,4 +1,8 @@
 import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
+// Required for async/await to work in tests
+import regeneratorRuntime from "regenerator-runtime";
+global.regeneratorRuntime = regeneratorRuntime;
```
guscost: ???

PR #139, opened 2019-02-01
time-travel reports

guscost:
Adds an Organization setting: `dashboard_week_scrubber` which when `true` will render a scrubber at the top of dashboards, allowing users to skip back/forwards a week at a time. THIS IS A TEMPORARY FEATURE.


src/components/dashboard/index.tsx
```diff
@@ -237,6 +246,31 @@ export function Dashboard({
         </AppSidebar>
         <AppPane>
           <AppBar
+            // TODO: Replace this with better report time navigation (like MixPanel)
+            rightSpan={settings && stringToBoolean(settings.dashboardWeekScrubber) ? <div style={{
+              flexGrow: 1,
```
1egoman: Can you move these styles into the stylesheet?

guscost: temporary feature!

beweinreich:
screen shot 2019-02-05 at 12 20 45 pm

Thoughts on makign it look like this...? It's a little less jarring.

beweinreich:
@guscost ^

src/components/dashboard/index.tsx
1egoman: Temporary features have a way of becoming permanent. ¯\_(ツ)_/¯

guscost: we're actively working on changing that reality, otherwise we won't be able to do incremental/temporary/experimental rollouts like this at all.

guscost:
Changed the buttons to be blue-on-white

PR #140, opened 2019-02-04
Only show spaces in search results

beweinreich:


src/components/explore/index.tsx
```diff
@@ -109,9 +109,12 @@ export function Explore({
   if (spaces.filters.search) {
     filteredSpaces = spaceFilter(filteredSpaces, spaces.filters.search);
   }
-  const spacesByName = filteredSpaces.sort(function(a, b) {
+  let spacesByName = filteredSpaces.sort(function(a, b) {
     return a.name.localeCompare(b.name);
   });
+  // only show "spaces" in search results
+  spacesByName.filter(space => space.spaceType == "space");
```
guscost: Could chain this, but it's a matter of taste and the comment helps.

beweinreich: Yeah I did chain it originally, but I thought the comment might be helpful. We can always remove later if it's superfluous.

PR #147, opened 2019-02-08
Small design fixes

guscost:
Fix the daily metrics chart, implement a new resize handler for the whole dashboard, and switch to a batch action for setting initial "space events" on all spaces at once.


src/components/explore-space-detail-daily-metrics-card/index.tsx
```diff
@@ -145,6 +145,14 @@ export class ExploreSpaceDetailDailyMetricsCard extends Component<any, any> {
           <CardBody className="explore-space-detail-daily-metrics-card-body">
             {calculatedData.state === 'COMPLETE' ? (() => {
               if (calculatedData.data.metrics.length > GRAPH_TYPE_TRANSITION_POINT_IN_DAYS) {
+                const data: any[] = [];
```
1egoman: Thoughts on simplifying this code to the below? (ie, get rid of the `.forEach(x => data.push(x))` at the end)
```
const data: any[] = calculatedData.data.metrics.sort(
  (a, b) => moment.utc(a.timestamp).valueOf() - moment.utc(b.timestamp).valueOf()
);
```

package.json
```diff
@@ -49,10 +49,10 @@
     "normalize.css": "^7.0.0",
     "popper.js": "^1.14.5",
     "qs": "^6.5.2",
-    "react": "^16.6.3",
+    "react": "^16.8.1",
```
1egoman: Nice!

guscost: Use hooks!

src/components/explore-space-detail-daily-metrics-card/index.tsx
guscost: This leads to the data getting longer by one each render (note the extra .push below), it needs to be a totally new reference I think, but there may be a better way to do so.

1egoman: Ah, I see what you mean. Yea ok, I guess what you have works, maybe you could do something like `const data: any[] = calculatedData.data.metrics.slice().sort(`? `.push` in a `.forEach` feels like an antipattern to me.

guscost: Beautiful, that's exactly what it needs. Thanks.

PR #148, opened 2019-02-11
Explore more

beweinreich:
Added in the hourly breakdown chart.


src/actions/route-transition/explore-space-trends.ts
```diff
@@ -296,3 +312,55 @@ export function calculateUtilization(space) {
     }));
   };
 }
+
+
+export function generateHourlyBreakdownEphemeralReport(space, startDate, endDate) {
```
1egoman: @beweinreich We had talked about moving this function into `src/helpers`. I don't think this function should really live in this file, it's used in both the component and in the action.

src/actions/route-transition/explore-space-trends.ts
```diff
@@ -296,3 +312,55 @@ export function calculateUtilization(space) {
...
+  return {
+    id: `rpt_${space.id}`,
+    type: 'HOURLY_BREAKDOWN',
+    name: 'Hourly Breakdown',
+    settings: {
+      "spaceId": space.id,
+      "timeRange": {
```
1egoman: FYI, these keys don't have to be quoted. (`timeRange` vs `"timeRange"`)

src/components/explore-space-trends/index.tsx
```diff
@@ -223,6 +228,21 @@ class ExploreSpaceTrends extends React.Component<any, any> {
                   chartWidth={this.state.width}
                 />
               </div>
+              <div className="explore-space-trends-item">
+                <Report
```
1egoman: As it is, `Report` isn't all that descriptive on its own. It might be worth writing a comment explaining that this report is an hourly breakdown report and is dynamically generated using values selected in the filter bar on this page?

src/components/explore-space-trends/_styles.scss
```diff
@@ -17,3 +17,7 @@
 .explore-space-trends-item {
   margin-bottom: 40px;
 }
+
+.explore-space-trends-item div {
```
1egoman: This is a bad thing to do. You're modifying the styles of the report so that its functionality can be "customized" to this situation, which means that now if the way this report works changes (ie, we maybe wrap the whole thing in a div for some reason) then this selector will no longer work as intended. It's this same kind of issue: https://densityio.atlassian.net/browse/DEN-7380

The correct solution here would be to do one of these things:
- Add a prop to the report that is something like `fullWidth` and then use that to set the style within the report itself.
- Alternatively, we could remove the `max-width` entirely from reports. I don't think that causes any issued, @guscost thoughts on this one?

src/components/explore-space-trends/index.tsx
1egoman: Also, you may want to consider some sort of loading state for this report. Currently, it continues to render the previous dataset while a new dataset is loading.

1egoman: One more thing - this report doesn't seem like it's clear across multiple weeks of data. See the below screenshot:

screen shot 2019-02-11 at 9 02 01 am

I'm not sure what we want to do here. I think that the "ideal" answer would be to modify the report so it can work across multiple weeks (ie, maybe adding the date before `Friday` if the time span is greater than a week or something)

src/actions/route-transition/explore-space-trends.ts
beweinreich: donezo

src/components/explore-space-trends/index.tsx
beweinreich: great catch. we should likely have the date in parenthesis if its greater than 7 days

beweinreich: @1egoman https://github.com/DensityCo/reports/pull/12

src/components/explore-space-trends/_styles.scss
beweinreich: https://github.com/DensityCo/reports/pull/11

guscost: Approved ^

src/components/explore-space-trends/index.tsx
beweinreich: Added in a loading indicator (and moved to its own component)

src/components/explore-space-trends/_styles.scss
beweinreich: gracias

src/components/explore-space-detail-hourly-breakdown-card/_styles.scss
```diff
@@ -0,0 +1,21 @@
+// Horizontally and vertically center errors and info messages in the 24 hour card.
```
1egoman: This isn't the 24 hour card!

beweinreich: Woops

PR #149, opened 2019-02-11
Added in intercom

beweinreich:


src/components/intercom/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  let intercomUser = {};
+  if (user.data && user.data.hasOwnProperty("id")) {
+    intercomUser = {
+      user_id: user.data.id,
+      email: user.data.email,
+      name: user.data.fullName
+    }
+  }
+  return (
+    <Intercom appID="sr9uzcmq" { ...intercomUser } />
```
1egoman: Should `appID` be an environment variable? Ie, something like `REACT_APP_INTERCOM_APP_ID`? I'm not sure if this needs to change depending on environment or whether it makes sense to use the same value in local testing / etc (local development or staging may skew the marketing team's analytics if they are being captured).

bredfield: Second'd

1egoman: @beweinreich If this is what we end up doing then make sure you update the environment variables [here](https://github.com/DensityCo/dashboard/blob/trunk/.circleci/config.yml#L80-L85) (staging) and [here](https://github.com/DensityCo/dashboard/blob/trunk/.circleci/config.yml#L111-L116) (production) so that they will make their way into the final build.

beweinreich: very cool. didnt know about that. will change it up

beweinreich:
@1egoman @bredfield ready for review.

src/components/intercom/index.tsx
```diff
@@ -16,9 +16,16 @@ export function IntercomDensity({
...
-  return (
-    <Intercom appID="sr9uzcmq" { ...intercomUser } />
-  );
+  if (process.env.REACT_APP_INTERCOM_APP_ID) {
+    return (
+      <Intercom appID={process.env.REACT_APP_INTERCOM_APP_ID} { ...intercomUser } />
+    )
+  } else {
+    return (
+      <div></div>
```
1egoman: @beweinreich Just FYI for the future - you can return `null` from a react component and it will render nothing!

PR #157, opened 2019-02-18
SSO integration

1egoman:
Adds a new button on the login page:
screen shot 2019-02-18 at 1 45 03 pm

This (at the moment) goes through a few redirects in order to log the user in once they have clicked the "log in with google" button:

```
http://localhost:3000/login-callback#access_token=TOKEN HERE

http://localhost:3000/login-callback#/login

http://localhost:3000/#/login-callback&access_token=TOKEN HERE

http://localhost:3000/#/dashboards/das_xxxxx
```


src/auth0.ts
```diff
@@ -0,0 +1,18 @@
+// Load Auth0 SDK
+import auth0 from 'auth0-js';
+
+const webAuth = new auth0.WebAuth({
+    // Auth0 Domain
+    domain: 'densityio.auth0.com',
+
+    // SPA settings
+    clientID: 'sisXLEhEmOcPaGToGHF0SQ1qvnCDRmbI',
```
1egoman: Should this value be an environment variable? I'm unsure if it's the sort of value we'd want to have easy access to modify or change depending on the local development / staging production.

src/components/login-callback/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import sessionTokenSet from "../../actions/session-token/set";
+import objectSnakeToCamel from "../../helpers/object-snake-to-camel";
+import unsafeNavigateToLandingPage from "../../helpers/unsafe-navigate-to-landing-page";
+import redirectAfterLogin from "../../actions/miscellaneous/redirect-after-login";
+
+import { accounts } from '../../client';
+
+
+export class LoginCallback extends React.Component<any, any> {
+  constructor(props) {
```
1egoman: FYI, the constructor can be omitted since it doesn't have any custom logic in it.

src/components/login-callback/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+    const vars = window.location.hash.split('&');
+    let token;
+    for (let i = 0; i < vars.length; i++) {
+        const pair = vars[i].split('=');
+        if (decodeURIComponent(pair[0]) == 'access_token') {
+            token = pair[1];
+        }
+    }
+
+    console.log(`token=${token}`);
```
1egoman: We should probably remove these logs before deploying.

src/components/login/_styles.scss
```diff
@@ -1,58 +1,123 @@
...
-// Wrapper for the login form
-.login-form-container {
-  text-align: center; // For forgot password link in the login form (goes to forgot password page)
+// Hacking the Google SSO button
+.login-submit-button.google {
+  margin-top: 0;
+}
+.login-submit-button.google button {
+  display: flex;
+  display: -webkit-box; /* OLD - iOS 6-, Safari 3.1-6 */
```
1egoman: We shouldn't need this, I'm pretty sure create-react-app (via postcss) should add these vendor prefixes.

src/index.tsx
```diff
@@ -64,10 +64,31 @@ import { collectionSpacesBatchSetEvents } from './actions/collection/spaces/set-
...
+// All the reducer and store code is in a separate file.
 import storeFactory from './store';
+import routeTransitionLoginCallback from "./actions/route-transition/login-callback";
 const store = storeFactory();
 
+
+// Auth0 strips everything following a hash for the callback URL and replaces it with the token information
+// so we can't use our usual hash-base routing - instead we check for our callback path and manually navigate
+// to the hash-based route, appending the Auth0 hash data to our own
+console.log(`pathname=${window.location.pathname}`);
```
1egoman: Another few logs we should probably remove prior to deploying.

src/components/login-callback/index.tsx
guscost: This component doesn't exist anymore

src/auth0.ts
1egoman: @cbigler ^

cbigler: The clientID is the same for all of the environments

src/components/login/_styles.scss
1egoman: @rjgrazioli said that these were needed, create react app apparently wasn't adding the prefixes automatically.

src/components/login/index.tsx
```diff
@@ -223,7 +234,7 @@ export class Login extends React.Component<any, any> {
           <div className="login-toast">
             <Toast
               type="danger"
-              title="Incorrect password"
+              title={this.state.errorTitle || 'Incorrect Password'}
```
guscost: Capitalization convention is changing here?

cbigler: Restored to original

PR #160, opened 2019-02-20
Maintain the date range when jumping between spaces

beweinreich:
Let me know if I need to clean up the `explore-space-trends.ts` page logic. But, essentially, this means that once you select a date range, you maintain that date range.

Apparently, this was driving customers nuts, since the date range would reset every time they clicked on a space.

### DIFF
--- a/src/actions/route-transition/explore-space-trends.ts
+++ b/src/actions/route-transition/explore-space-trends.ts
@@ -82,13 +82,23 @@ export default function routeTransitionExploreSpaceTrends(id) {
     dispatch(collectionSpacesSetDefaultTimeRange(selectedSpace));
     dispatch(collectionSpacesSet(spaces));
 
+    let startDate, endDate;
+    let state = getState()
+    if(state.spaces.filters.startDate != null && state.spaces.filters.endDate != null) {
+      startDate = state.spaces.filters.startDate
+      endDate = state.spaces.filters.endDate
+    } else {
+      startDate = formatInISOTimeAtSpace(getCurrentLocalTimeAtSpace(selectedSpace).subtract(1, 'week').startOf('week'), selectedSpace);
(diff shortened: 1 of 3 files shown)


src/actions/route-transition/explore-space-trends.ts
```diff
@@ -82,13 +82,23 @@ export default function routeTransitionExploreSpaceTrends(id) {
     dispatch(collectionSpacesSetDefaultTimeRange(selectedSpace));
     dispatch(collectionSpacesSet(spaces));
 
+    let startDate, endDate;
+    let state = getState()
+    if(state.spaces.filters.startDate != null && state.spaces.filters.endDate != null) {
```
1egoman: Can you make this `if (`? Sorry, one of those things that drives me crazy 😄

src/reducers/spaces/index.ts
```diff
@@ -205,26 +205,30 @@ export default function spaces(state=initialState, action) {
...
-        // data
-        startDate: formatInISOTime(
-          getCurrentLocalTimeAtSpace(action.space).subtract(1, 'week').startOf('week')
-        ),
-        endDate: formatInISOTime(
-          getCurrentLocalTimeAtSpace(action.space).subtract(1, 'week').endOf('week')
-        ),
-      },
-    };
+    if(state.filters.date != null && state.filters.startDate != null && state.filters.endDate != null) {
```
1egoman: here too

src/actions/route-transition/explore-space-trends.ts
beweinreich: yep np problem

src/reducers/spaces/index.ts
beweinreich: got it

PR #166, opened 2019-02-27
Filter by ancestry

beweinreich:
This changes the search field so that it uses the ancestry part of the space object to filter up the hierarchy. Gives you a better sense as to where you are when searching.

screen shot 2019-02-26 at 11 24 25 pm

### DIFF
--- a/src/components/explore/index.tsx
+++ b/src/components/explore/index.tsx
@@ -1,6 +1,7 @@
 import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
+import fuzzy from 'fuzzy';
 
 import {
   AppBar,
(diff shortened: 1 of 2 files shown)


src/components/explore/index.tsx
```diff
@@ -126,18 +127,35 @@ export function Explore({
...
-  let spacesByName = filteredSpaces.sort(function(a, b) {
-    return a.name.localeCompare(b.name);
-  });
-  // only show "spaces" in search results
-  spacesByName.filter(space => space.spaceType == "space");
 
   const spaceTree = convertSpacesToSpaceTree(filteredSpaces)
-  const spaceList = spaces.filters.search ? spacesByName : spaceTree;
-
+  const spaceList = spaceTree;
```
1egoman: Can this line and the one above it be turned into `const spaceList = convertSpacesToSpaceTree(filteredSpaces)`?

beweinreich: Yep. Good call.

PR #169, opened 2019-03-06
Feat: account admin

guscost:


src/actions/route-transition/admin-user-management.ts
```diff
@@ -0,0 +1,12 @@
+export const ROUTE_TRANSITION_ADMIN_USER_MANAGEMENT = 'ROUTE_TRANSITION_ADMIN_USER_MANAGEMENT';
+
+import collectionUsersSet from '../collection/users/set';
+import { accounts } from '../../client';
+
+export default function routeTransitionAdminUserManagement() {
+  return async dispatch => {
+    dispatch({ type: ROUTE_TRANSITION_ADMIN_USER_MANAGEMENT });
+    const users = await accounts.users.list();
```
1egoman: Should there be a `try {} catch {}` around this call?

src/actions/toasts/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default function showToast({
+  text,
+  title = undefined as string | undefined,
+  icon = undefined as any | undefined,
+  type = undefined as string | undefined,
+  timeout = 2000
+}) {
+  return dispatch => {
+    const id = Math.random().toString();
```
1egoman: I'd prefer we use uuids unless there's a reason in particular that this way is better.

src/components/admin-user-management/index.tsx
```diff
@@ -0,0 +1,317 @@
...
+    {/* If the "add user" modal is visible, render it above the view */}
+    {activeModal.name === 'MODAL_ADMIN_USER_ADD' ? (
+      <Modal
+        visible={activeModal.visible}
+        width={480}
+        onBlur={onCancelAddUser}
+        onEscape={onCancelAddUser}
+      >
+        <div style={{display: 'flex', flexDirection: 'column'}}>
+          {/* <div style={{marginTop: -64}}>
```
1egoman: Dead code

src/components/admin/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  AppBarTitle,
+  AppFrame,
+  AppPane,
+} from '@density/ui';
+
+import Toaster from '../toaster';
+import AppBarSubnav, { AppBarSubnavLink } from '../app-bar-subnav';
+import AdminUserManagement from '../admin-user-management';
+import can from '../../helpers/permissions';
+// import AdminDeveloper from '../admin-developer';
```
1egoman: Dead code

src/components/app-navbar/index.tsx
```diff
@@ -7,156 +7,210 @@ import {
...
+      </a>
+    </li>
+  );
+}
+
+
+// Define the real app navbar here
+export default function AppNavbar({page, user, settings}) {
+
+  const showAdminMenu = can(user, 'developer_tools_manage') ||
```
1egoman: I like this `can` function a lot!

src/components/app-navbar/index.tsx
```diff
@@ -7,156 +7,210 @@ import {
...
+            this.selectBoxValueRef.blur();
+            this.onMenuBlur();
+          }
+        }}
+      >
+        {header}
       </div>
-    );
+      <nav className={classnames('app-navbar-menu-items', {opened})}>
+        <AppNavbarMenuContext.Provider value={this}>
```
1egoman: I tend to not like `this` as the context value since it's not apparent what keys on `this` are being used. After analyzing the below code, maybe this should be `<AppNavbarMenuContext.Provider value={{onMenuFocus: this.onMenuFocus, onMenuBlur: this.onMenuBlur}}>`? I think it's functionally identical but it makes it clear what attributes in `this` are being used.

src/actions/route-transition/admin-user-management.ts
guscost: Yes, but there are a whole bunch of requests missing this right now. We should do this as an across-the-board update.

src/actions/toasts/index.ts
guscost: Ah hah hah ok.

src/components/admin-user-management/index.tsx
guscost: I want to keep this as an example until we update the "report detail view" modal

src/components/admin/index.tsx
guscost: Resolved in #172

src/components/app-navbar/index.tsx
guscost: Me too!

src/components/app-navbar/index.tsx
guscost: Resolved in #172

PR #175, opened 2019-03-12
Shouldn't say "Average" when it's less than a week

beweinreich:
The hourly peak chart shouldn't say "Average" if its just for a week.

### DIFF
--- a/src/helpers/multi-week-selection/index.ts
+++ b/src/helpers/multi-week-selection/index.ts
@@ -1,0 +1,11 @@
+import moment from 'moment';
+
+export default function isMultiWeekSelection(startDate, endDate) {
+	let multiWeekSelection = false;
+	const startDateMoment = moment(startDate);
+	const endDateMoment = moment(endDate);
+	if (endDateMoment.diff(startDateMoment, 'days') > 7) {
+		multiWeekSelection = true;
+	}
+	return multiWeekSelection;
+}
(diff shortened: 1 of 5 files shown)


src/helpers/multi-week-selection/index.ts
```diff
@@ -0,0 +1,11 @@
+import moment from 'moment';
+
+export default function isMultiWeekSelection(startDate, endDate) {
+	let multiWeekSelection = false;
+	const startDateMoment = moment(startDate);
```
1egoman: Keep in mind `moment()` uses the local time zone of the user for this calculation. I don't know if that's what you intended (I suspect not).

src/helpers/multi-week-selection/index.ts
```diff
@@ -0,0 +1,11 @@
...
+	const endDateMoment = moment(endDate);
+	if (endDateMoment.diff(startDateMoment, 'days') > 7) {
+		multiWeekSelection = true;
+	}
+	return multiWeekSelection;
```
1egoman: You could probably simplify this to `return endDateMoment.diff(startDateMoment, 'days') > 7`

beweinreich: good call.

src/helpers/multi-week-selection/index.ts
beweinreich: thats fine.

PR #176, opened 2019-03-14
feature: debug impersonation

guscost:
Adds an impersonation function for debugging. Also replaces all AJAX code with Axios.


guscost:
These changes are in #167

PR #180, opened 2019-03-15
code: read custom week start from org settings

guscost:


src/actions/route-transition/dashboard-detail.ts
```diff
@@ -36,13 +37,19 @@ function loadDashboardAndReports(id) {
     const dashboardDate = getState().miscellaneous.dashboardDate;
     let dashboardSelectionPromise;
 
+    // TODO: get organization settings from another reducer 
+    const user = (await accounts().get('/users/me')).data;
```
1egoman: This can be (I think) `const user = getState().users.data;`

guscost: it's `user` and i think i tried that, it's not guaranteed to exist since this happens in parallel with the user fetch, i think

1egoman: Hmm ok

guscost: Fixed it!

PR #183, opened 2019-03-18
Remove user input for pen-test findings

beweinreich:


beweinreich:
> Looks good. We should also remove the email field from the encoded data passed to this page too while we're at it.

Hmm I'm not seeing it? We can punt for now, but I'll create a JIRA once I know what ya mean.

1egoman:
@beweinreich What I mean is this: https://github.com/DensityCo/dashboard/pull/183/files#diff-7676dde7da5c1da99d3bfcc6a488f4b1R26

In this data, we have encoded an `email` value. You removed the only place where it was used, and therefore, we can remove the field from the data that the core api generates.

1egoman:
@beweinreich What I mean is this: https://github.com/DensityCo/dashboard/pull/183/files#diff-7676dde7da5c1da99d3bfcc6a488f4b1R26

In this data, we have encoded an `email` value. You removed the only place where it was used, and therefore, we can remove the field from the data that the core api generates.

beweinreich:
Got it. Will remove that as well

PR #189, opened 2019-03-21
Integrations page - Robin Integration

beweinreich:
Introduces an Admin Integrations page.

Right now the only Integration we have is Robin. However, this page is ready to handle multiple integrations. They will be pulled down dynamically from the server.

There are `Service` and `ServiceAuthorization` objects.

----
Service object:

Services have a `name` and `display_name` and `category`. 
Category is `Chat` or `Room Booking` 

`name` is the name of the service, like `google_calendar`
`display_name` is the print friendly version like `Google Calendar`

----
ServiceAuthorization

The service authorization object stores the credentials, last_time it was synced, and whether the service is the default service for the category.

`credentials` is JSON blog that is unique per service
`default` whether its the default service
`last_sync` when it was refreshed (incase we need to refresh on the server).
`user` the user who added the service authorization

----

The admin page fetches a list of `Service` objects that have a `service_authorization` object inside of it. If the `service_authorization` exists, it will show an edit button, otherwise it shows an Activate button.

You can also destroy the 

Screen Shot 2019-03-21 at 1 39 05 PM

Screen Shot 2019-03-21 at 1 39 18 PM
Screen Shot 2019-03-21 at 1 39 10 PM


src/helpers/websocket-event-pusher/index.ts
```diff
@@ -42,6 +42,8 @@ export default class WebsocketEventPusher extends EventEmitter {
   }
 
   async connect(iteration=0) {
+    return;
```
guscost: ???

src/reducers/spaces/index.ts
```diff
@@ -186,7 +186,6 @@ export default function spaces(state=initialState, action) {
         timeSegmentGroupId: newTimeSegmentGroupId
       }
     };
-  case ROUTE_TRANSITION_EXPLORE_SPACE_TRENDS:
```
guscost: ???

src/actions/collection/service-authorizations/create.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (serviceName == "robin") {
+      request_body = {
+        "robin_access_token": serviceAuthorization.credentials.robinAccessToken,
+        "robin_organization_id": serviceAuthorization.credentials.robinOrganizationId,
+      }
+    }
+
+    try {
+      const response = await core().post(`/integrations/${serviceName}/`, request_body);
+      dispatch(integrationServicesList());
```
1egoman: You want this dispatch to not be in the `try`, because if it is, then any errors that your component has while rendering will be caught and you won't see the stack trace. You want to do something like this:
```
    let response;
    try {
      response = await core().post(`/integrations/${serviceName}/`, request_body);
    } catch (err) {
        dispatch(collectionServiceAuthorizationsError(err));
        return false;
    }
    dispatch(integrationServicesList());
```

src/actions/collection/service-authorizations/destroy.ts
```diff
@@ -0,0 +1,20 @@
...
+import core from '../../../client/core';
+
+export const COLLECTION_SERVICE_AUTHORIZATIONS_DESTROY = 'COLLECTION_SERVICE_AUTHORIZATION_DESTROY';
+
+export default function collectionServiceAuthorizationDestroy(serviceAuthorizationId) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_SERVICE_AUTHORIZATIONS_DESTROY, serviceAuthorizationId });
+
+    try {
+      await core().delete(`/integrations/service_authorizations/${serviceAuthorizationId}`);
```
1egoman: Same thing with the dispatch being in the `try` here

src/actions/collection/service-authorizations/update.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const COLLECTION_SERVICE_AUTHORIZATIONS_UPDATE = 'COLLECTION_SERVICE_AUTHORIZATIONS_UPDATE';
+
+export default function collectionServiceAuthorizationUpdate(serviceName, serviceAuthorization) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_SERVICE_AUTHORIZATIONS_UPDATE, serviceAuthorization });
+
+    let request_body = {};
+    if (serviceName == "robin") {
+      request_body = {
```
1egoman: camel cased variable names please!

src/actions/integrations/space_mapping.ts
```diff
@@ -0,0 +1,36 @@
...
+        });
+      }
+
+      // refetch spaces for right now
+      // so that the spaceMappings in the space object is correct
+      const spaces = await core().get(`/spaces`);
+
+      dispatch(collectionSpacesSet(spaces.data.results));
+
+    } catch (err) {
```
1egoman: Do you want to do anything with this error here?

src/components/admin-integrations-robin-create-modal/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import Modal from '../modal';
+import { CancelLink } from '../dialogger';
+
+import FormLabel from '../form-label';
+
+export default class IntegrationsRobinCreateModal extends React.Component<any, any> {
+  constructor(props) {
+    super(props);
+    this.state = {
```
1egoman: It's slightly preferred these days to use this form:
```
export default class IntegrationsRobinCreateModal extends React.Component<any, any> {
  state = {
    robinAccessToken: '',
    robinOrganizationId: '',
  }

  ...

}
```

src/components/admin-integrations-robin-update-modal/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+import FormLabel from '../form-label';
+import Modal from '../modal';
+import { CancelLink } from '../dialogger';
+
+export default class IntegrationsRobinUpdateModal extends React.Component<any, any> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      robinAccessToken: this.props.serviceAuthorization.credentials.robinAccessToken,
```
1egoman: Keep in mind here that if you update the value of the `serviceAuthorization` prop, then it's not going to make it's way in the state. This is usually an anipattern in react but when I have to do it I try to name the prop something like `initialServiceAuthorization` to make it clear that changing the prop won't cause the component to update.

src/reducers/spaces/index.ts
beweinreich: whooops

src/helpers/websocket-event-pusher/index.ts
beweinreich: ha sorry. the websockets were driving me nuts! ill remove that. i should just make that into an env var

beweinreich: glad you caught it!

src/actions/collection/service-authorizations/create.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const COLLECTION_SERVICE_AUTHORIZATIONS_CREATE = 'COLLECTION_SERVICE_AUTHORIZATIONS_CREATE';
+
+export default function collectionServiceAuthorizationCreate(serviceName, serviceAuthorization) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_SERVICE_AUTHORIZATIONS_CREATE });
+
+    let request_body = {}
+    if (serviceName == "robin") {
+      request_body = {
```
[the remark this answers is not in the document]
beweinreich: Oof— switching between python and JS too much! Good catch

src/actions/collection/service-authorizations/create.ts
beweinreich: ah yeah you told me about this! okay fixing.

src/actions/integrations/space_mapping.ts
beweinreich: This isn't really "complete" yet. It's for the next PR which will include the meetings page. I'll handle that in the next PR. I can remove this whole thing, but its just a pain.

src/components/admin-integrations-robin-create-modal/index.tsx
beweinreich: I like it. Much cleaner.

1egoman:
Oh actually one more thing @beweinreich - can you rename `src/actions/integrations/space_mapping.js` to `src/actions/integrations/space-mapping.js`? That's the naming convention we've been using.

beweinreich:
> Oh actually one more thing @beweinreich - can you rename `src/actions/integrations/space_mapping.js` to `src/actions/integrations/space-mapping.js`? That's the naming convention we've been using.

Done.

PR #195, opened 2019-04-04
Space management breadcrumb

1egoman:
This pull request has the beginnings of the space management project contained within it.

- Adds a new page, routable at `#/admin/locations`. When loaded, it makes a request to the server to load all spaces (if all spaces are already loaded, the request is not made and the spaces that are already in the store are used instead)
- Adds a component that renders a breadcrumb relating to the currently selected space using the space's `ancestry` value.
- Currently, it renders a `<ul>` containing links to all direct children of the space that is currently selected. The interface that goes here is coming in a future pull request.


src/types.ts
```diff
@@ -18,6 +18,7 @@ export type DensitySpace = {
 	currentCount: number,
 	capacity: number,
 	createdAt: string,
+  ancestry: Array<DensitySpaceAncestryItem>,
```
guscost: got some weird tab behavior here?

src/reducers/spaces/index.ts
```diff
@@ -47,6 +48,7 @@ const DATA_DURATION_WEEK = 'DATA_DURATION_WEEK';
       // DATA_DURATION_MONTH = 'DATA_DURATION_MONTH';
 
 const initialState = {
+  view: 'LOADING',
```
guscost: It bothers me that the dashboard uses `state` for this concept unlike almost everything else, and I didn't realize it until just now (when we probably shouldn't touch that stuff again for a while). This and other little inconsistencies with the collection action naming and such should all go in some sort of "fix these mildly annoying consistency pet peeves that gus complains about" project in our next engineering to-do list.

src/components/admin-locations/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+            </AppBarSection>
+            <AppBarSection>
+              <span className={styles.leftButton}>
+                <Button disabled type="primary">
+                  <Skeleton width={96} color={colorVariables.grayLight} />
+                </Button>
+              </span>
+              <span className={styles.rightButton}>
+                <Button disabled type="primary">
+                  <Skeleton width={96} color={colorVariables.grayLight} />
```
guscost: ☠️ 👍

src/components/admin-locations-breadcrumb/index.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { Fragment } from 'react';
+import styles from './styles.module.scss';
+
+import { Icons } from '@density/ui';
+
+export default function Breadcrumb({ space }) {
```
guscost: Shouldn't this be AdminLocationsBreadcrumb (another component name consistency grumbling goes here)

src/components/.template/index.tsx
```diff
@@ -4,7 +4,9 @@ import styles from './styles.module.scss';
 export default function %COMPONENTUPPERCAMEL%({
   name,
 }) {
-  return <div className={styles.%COMPONENTCAMEL%}>
-    {name ? `Hello ${name}` : 'Hello World!'}
-  </div>;
+  return (
```
guscost: Sure, whatever, let's waste a couple lines of code for nice clean parens 😛

src/actions/route-transition/admin-locations.ts
```diff
@@ -0,0 +1,44 @@
...
+    dispatch({
+      type: ROUTE_TRANSITION_ADMIN_LOCATIONS,
+      parentSpaceId,
+      setLoading: shouldLoadSpaces,
+    });
+
+    if (shouldLoadSpaces) {
+      let spaces;
+      try {
+        const rawSpaces = await fetchAllPages(async page => {
```
guscost: Should we abstract this chunk? It's used a lot and this is a nice version of it.

1egoman: Haha I actually started going down the path to abstract this this morning! I had made an action called something like `collectionSpacesFetchAll` that had this logic in it and would be called from other actions. Let's talk about it and see how we can get that change in a separate PR.

src/components/admin-locations-breadcrumb/index.tsx
1egoman: Yep, will do

src/reducers/spaces/index.ts
1egoman: Sounds good, some consistency would be helpful!

src/actions/route-transition/admin-locations.ts
guscost: 👍

PR #196, opened 2019-04-05
Add sibling space selector to the locations breadcrumb

1egoman:
Add breadcrumb space selector, which uses the list of loaded spaces to display all siblings and allow navigation to them.

Rough example:

### DIFF
--- a/src/components/admin-locations-breadcrumb/index.tsx
+++ b/src/components/admin-locations-breadcrumb/index.tsx
@@ -1,9 +1,14 @@
-import React, { Fragment } from 'react';
+import React, { Component, Fragment } from 'react';
+import classnames from 'classnames';
 import styles from './styles.module.scss';
 
-import { Icons } from '@density/ui';
+import { DensitySpace } from '../../types';
 
-export default function AdminLocationsBreadcrumb({ space }) {
+import { InputBox, Icons } from '@density/ui';
+
+import filterCollection from '../../helpers/filter-collection/index';
+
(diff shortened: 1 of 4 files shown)


src/components/admin-locations-breadcrumb/index.tsx
```diff
@@ -37,3 +44,127 @@ export default function AdminLocationsBreadcrumb({ space }) {
...
+              })}
+              onKeyDown={e => {
+                if (e.key === 'ArrowDown') {
+                  if (activeItemIndex === null) {
+                    this.setState({activeItemIndex: 0});
+                  } else if (
+                    typeof activeItemIndex === 'number' &&
+                    (activeItemIndex as any) < items.length-1
+                  ) {
+                    this.setState({activeItemIndex: (activeItemIndex as any) + 1});
```
1egoman: I have a number of typescript issues with `activeItemIndex` since its type is `number | null`. Not sure how to get around these at the typescript level (ideas welcome!) but instead I checked the type at runtime above in the `if` and then casted to any. (casting to `number` doesn't work here unfortunately because of the union type 😦 )

beweinreich:
That was quick

src/components/admin-locations-breadcrumb/index.tsx
guscost: Per conversation, this can be -1 for now. Slightly surprising but a lot cleaner than casting (and better than giving up and making it `any` across the board)

PR #198, opened 2019-04-09
User view design

rjgrazioli:


src/components/admin-user-management-detail/index.tsx
```diff
@@ -161,91 +161,107 @@ export class AdminUserManagementDetail extends Component<AdminUserManagementDeta
...
+                    onChangeActive={spaceFilteringActive => this.setState({spaceFilteringActive})}
+                    selectedSpaceIds={this.state.spaceIds}
+                    onChange={spaceIds => this.setState({spaceIds})}
+                    height={600}
+                  />
+                </div>
+              </div>
+
+              <div className={styles.adminUserManagementDetailSection}>
+                <div className={`${styles.adminUserManagementDetailCard} ${styles.adminUserManagementDetailCardDanger}`}>
```
1egoman: Can you make this use `classnames`?

1egoman:
I'd also like @guscost to look at this too if possible

src/components/admin-user-management-detail/index.tsx
```diff
@@ -161,91 +161,107 @@ export class AdminUserManagementDetail extends Component<AdminUserManagementDeta
...
-                    <span className={styles.adminUserManagementUserInfoEmail}>{selectedUser.email}</span>
-                    <div>
-                      <span className={styles.adminUserManagementUserInfoRoleTag}>
+
+                      <div className={styles.adminUserManagementUserInfoRoleTag}>
                         {ROLE_INFO[selectedUser.role].label}
-                      </span>
+                      </div>
+
+                      <dl className={styles.adminUserManagementUserInfoDetails}>
```
guscost: What's this thing?

src/components/admin-user-management-detail/index.tsx
```diff
@@ -161,91 +161,107 @@ export class AdminUserManagementDetail extends Component<AdminUserManagementDeta
...
+                    onChange={spaceIds => this.setState({spaceIds})}
+                    height={600}
+                  />
+                </div>
+              </div>
+
+              <div className={styles.adminUserManagementDetailSection}>
+                <div className={`${styles.adminUserManagementDetailCard} ${styles.adminUserManagementDetailCardDanger}`}>
+                  <AppBar>
+                    <AppBarTitle>Danger Zone</AppBarTitle>
```
guscost: The descender on the "g" is cut off. Also kinda lukewarm about calling a thing "Danger Zone" since it's been a recurring nostalgia joke on Archer FX, and may have been ruined for general use.

src/components/admin-user-management-detail/index.tsx
rjgrazioli: `<dl>` is a "description" list. it's probably a little more semantic than an `<li>` in this case, but also doesn't come with bullets by default and remove an override step.

src/components/admin-user-management-detail/index.tsx
rjgrazioli: I addressed this here https://github.com/DensityCo/ui/pull/113

Regarding "Danger Zone". I'm just copying GitHub. I'm not sure what Archer's audience is. I used to watch, but have no idea what you're talking about. I'm open to other alternatives. But this always seemed like, fun, approachable copy to me.

src/components/admin-user-management-detail/index.tsx
rjgrazioli: done

PR #199, opened 2019-04-09
Space management detail pages

1egoman:
This pull request adds detail pages within the space management ui for campuses, buildings, and floors.

The styling is still a bit wonky though.

### DIFF
--- a/src/components/admin-locations-building-detail/index.tsx
+++ b/src/components/admin-locations-building-detail/index.tsx
@@ -15,10 +15,10 @@ import {
   Icons,
 } from '@density/ui';
 
-function AdminLocationsBuildingDetailSpaceList({ spaces }) {
+function SpaceList({ spaces, renderedSpaces }) {
   return (
     <div className={styles.spaceList}>
-      <ListView data={spaces}>
+      <ListView data={renderedSpaces}>
         <ListViewColumn
           title="Info"
           template={item => (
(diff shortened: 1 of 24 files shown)


src/components/admin-locations-building-detail/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+          href={item => `#/admin/locations/${item.id}`}
+        />
+        <ListViewColumn
+          title="Capacity"
+          template={item => 'HARDCODED'}
+          href={item => `#/admin/locations/${item.id}`}
+        />
+        <ListViewColumn
+          title="DPUs"
+          template={item => 'HARDCODED'}
```
1egoman: All these rows have values of `HARDCODED` to make it really obvious that these fields either need to come from the api or they need to be calculated somehow.

src/components/admin-locations-building-detail/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+export default function AdminLocationsBuildingDetail({ spaces, selectedSpace }) {
+  const visibleSpaces = spaces.data.filter(s => s.parentId === selectedSpace.id);
+  const floors = visibleSpaces.filter(space => space.spaceType === 'floor');
+  const spacesInEachFloor = floors.map(floor => spaces.data.filter(space => space.parentId === floor.id));
+  const spacesNotInFloor = visibleSpaces.filter(space => floors.find(floor => space.parentId === floor.id));
+
+  return (
+    <AppFrame>
+      <AppSidebar visible>
+        TODO: Waiting on mockups
```
1egoman: Rob and I decided to wait on doing the left panes of each of these pages also for now

PR #200, opened 2019-04-10
Only track non-impersonated users

beweinreich:
This checks localstorage to see if a user is currently in impersonation mode and then doesn't track them in mixpanel


beweinreich:
@1egoman gimme a 👍 👎

PR #202, opened 2019-04-10
Don't run trends calculations when start time is after end time

1egoman:


beweinreich:
ezgif-4-5879ca336511

@1egoman I think we're good to go here.

PR #203, opened 2019-04-10
Robin / Teem updates

beweinreich:
- Updated Integrations page to handle new Teem service
- "Make default" to switch which service is default
- Delete buttons for Robin / Teem / Slack
- Show teem spaces in meetings page
- Genericized the room booking calls to be DRY

Screen Shot 2019-04-11 at 12 14 46 PM


src/actions/collection/service-authorizations/update.ts
```diff
@@ -19,13 +19,44 @@ export default function collectionServiceAuthorizationUpdate(serviceName, servic
...
       return response;
+    }
+  }
+}
+
+export function collectionServiceAuthorizationMakeDefault(serviceAuthorizationId) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_SERVICE_AUTHORIZATIONS_UPDATE });
+
+    let requestBody = {
```
1egoman: This object could probably be put inline into the request, like:
```
await core().put(`/integrations/service_authorizations/${serviceAuthorizationId}/`, {'default': true);
```

(note it does have to be quoted because `default` is a keyword in javascript)

src/actions/route-transition/admin-integrations-slack.ts
```diff
@@ -0,0 +1,17 @@
...
+import core from '../../client/core';
+
+import integrationServicesList from '../integrations/services';
+
+export const ROUTE_TRANSITION_ADMIN_INTEGRATIONS = 'ROUTE_TRANSITION_ADMIN_INTEGRATIONS';
+
+export default function routeTransitionAdminIntegrationsSlack(code) {
+  return async dispatch => {
+    dispatch({ type: ROUTE_TRANSITION_ADMIN_INTEGRATIONS });
+    await core().get('integrations/slack/auth', { params: { code: code } });
```
1egoman: This call should probably have a `try`/`catch` around it to handle errors if they happen?

src/actions/route-transition/explore-space-meetings.ts
```diff
@@ -119,16 +119,21 @@ export default function routeTransitionExploreSpaceMeeting(id) {
       }
     })();
 
-    // Load robin spaces if robin is active
-    if (roomBookingDefaultService && roomBookingDefaultService.name === 'robin') {
-      let robinSpaces;
+    // Load room booking spaces if room booking service is active
+    let roomBookingProvider;
+    if (roomBookingDefaultService && ['robin', 'teem'].includes(roomBookingDefaultService.name)) {
```
1egoman: Would it be better for the second expression in the if statement check to see the service type is something like `room_booking`? I seem to remember a field like that existing on a service.

src/components/admin-integrations/index.tsx
```diff
@@ -102,54 +106,75 @@ export function AdminIntegrations({
...
+          <ListViewColumn title="Name" template={item => (
+            <span className={styles.adminIntegrationsListviewValue}><strong>{item.displayName}</strong></span>
+          )} />
+          <ListViewColumn title="Added By" template={item => item.serviceAuthorization.id != null ? (
+            <span className={styles.adminIntegrationsListviewValue}>{item.serviceAuthorization.user.fullName}</span>) : null
+          } />
+          <ListViewColumn flexGrow={1} flexShrink={1} />
+          <ListViewColumn
+          template={item => item.serviceAuthorization.id == null ? <ListViewClickableLink>Activate</ListViewClickableLink> : null }
+          onClick={item => item.serviceAuthorization.id == null ? window.location.href = `https://slack.com/oauth/authorize?client_id=${process.env.REACT_APP_SLACK_CLIENT_ID}&scope=channels:read chat:write:bot&redirect_uri=${process.env.REACT_APP_SLACK_REDIRECT_URL}` : null } />
```
1egoman: IMO, this should probably be converted into an `if`. Having an assignment in a ternary like this is a bit strange.

src/actions/route-transition/explore-space-meetings.ts
1egoman: Although I think I understand why you did this now - if you added a new room booking service, you don't have the logic to fetch the spaces for it.

src/actions/collection/service-authorizations/update.ts
beweinreich: 👍

src/actions/route-transition/admin-integrations-slack.ts
beweinreich: yep

src/actions/route-transition/explore-space-meetings.ts
beweinreich: Yeah it's a weird balance. Like you can "dynamically" add new services in the backend, but really the front-end has to be ready for it.

src/components/admin-integrations/index.tsx
beweinreich: ugh c'mon jordo! okay ill fix

PR #206, opened 2019-04-11
Enhancement/ui cleanup

rjgrazioli:
Just polished up a few things to more closely resemble the latest designs. I'll probably hit up UI next to adjust sidebar widths and stuff.


src/components/admin-integrations/index.tsx
```diff
@@ -32,6 +32,8 @@ import collectionServiceAuthorizationCreate from '../../actions/collection/servi
 import collectionServiceAuthorizationUpdate from '../../actions/collection/service-authorizations/update';
 import collectionServiceAuthorizationDestroy from '../../actions/collection/service-authorizations/destroy';
 
+const DASHBOARD_BACKGROUND = '#FAFAFA';
```
1egoman: I'd prefer this to be defined in one place and imported in versus in each file that it's used. Thoughts on that?

rjgrazioli: yeah, for sure. we should just be able to reference the variables in UI, right?

rjgrazioli: `import colorVariables from '@density/ui/variables/colors.json'`

guscost: yes

PR #210, opened 2019-04-16
Space management edit page

1egoman:
Adds a page `#/admin/locations/:spaceId/edit` that shows a number of what we are calling "modules" that can be used to edit the space.

There is a separate component for each space type that is rendered on this page, and each separate component renders a different subset of modules. Each module accepts a `formState` prop that contains the total state of the form and a `onChangeField` prop that accepts the key in `formState` to modify and the new value to set it to.

There is currently no way to navigate to the edit page via the interface, you must manually go to the page. Also, submitting the form does nothing yet either.

---

Screen Shot 2019-04-16 at 12 40 11 PM


rjgrazioli:
Everything works as expected. 2 things

- [ ] I think the rent field should also be a number field
- [ ] placeholders

src/actions/route-transition/admin-locations-edit.ts
```diff
@@ -0,0 +1,35 @@
...
+      type: ROUTE_TRANSITION_ADMIN_LOCATIONS_EDIT,
+      spaceId,
+      setLoading: !space,
+    });
+
+    if (!space) {
+      let space;
+      try {
+        const response = await core().get(`/spaces/${spaceId}`);
+        space = objectSnakeToCamel<DensitySpace>(response.data);
```
guscost: Tidy and type-safe, love it

src/components/admin-locations-edit/index.tsx
```diff
@@ -0,0 +1,265 @@
+import React, { ReactNode, Fragment, Component } from 'react';
+import { connect } from 'react-redux';
+import styles from './styles.module.scss';
+
+import { DensitySpace } from '../../types';
+
+import {
+  AdminLocationsDetailModulesGeneralInfo,
```
guscost: These first-version names are 💯 👏 🔥 :chef:, so easy to follow

PR #212, opened 2019-04-17
Finish up address selector

1egoman:
- Starts empty and zoomed out to the whole world
- A user types an address into the search box, which shows a number of autocomplete suggestions. The user selects one of these suggestions.
- The map "flies to" this location.
- To clear the map, click the "x" in the address box. This workflow can be repeated to select a new address.

What's left:
- Map styling with mapbox's style tool

**NOTE**: The base of this PR is right now `space-management-modules` (so that the diff is easier to read), once https://github.com/DensityCo/dashboard/pull/210 is merged I will switch the base of this PR to `trunk`.


rjgrazioli:
Ok, I don't see any issues in the application. Everything works to spec and behaves the way I'd expect. I think there are a few things to consider from a design standpoint outside of the aesthetic changes that we can make later.

*Feel Strongly About:*
- [ ] Increase the width of the search box and drop down to 440px to accommodate more of the address
- [ ] Add the ability to re-center and re-zoom the map if you've moved or zoomed the map

*Uncertain About:*
- [ ] Add zoom controls instead of scroll-to-zoom to prevent accidental zoom

In general, I think moving the pin around makes sense and is a nice way to drop a pin where you want it. I think it'll go a long way from better building representation on campuses.

1egoman:
> Add the ability to re-center and re-zoom the map if you've moved or zoomed the map
>  Add zoom controls instead of scroll-to-zoom to prevent accidental zoom

If these changes are desired, we can make a separate PR for them.

PR #225, opened 2019-04-30
added in hotjar react package and tracking only if not in dev mode an…

beweinreich:
added in hotjar react package and tracking only if not in dev mode and not impersonating

### DIFF
--- a/src/helpers/analytics-user-reducer-enhancer/index.ts
+++ b/src/helpers/analytics-user-reducer-enhancer/index.ts
@@ -1,4 +1,6 @@
 import mixpanelInitialize from '../mixpanel-initialize/index';
+import { hotjar } from 'react-hotjar';
+
 
 function isUserImpersonating() {
   try {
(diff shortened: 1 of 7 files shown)


src/helpers/analytics-user-reducer-enhancer/index.ts
```diff
@@ -9,12 +11,17 @@ function isUserImpersonating() {
...
 
-export default function mixpanelUserReducerEnhancer(reducer) {
+export default function analyticsUserReducerEnhancer(reducer) {
   return (state, props) => {
     const result = reducer(state, props);
 
     if (!isUserImpersonating()) {
       // Not impersonating, so tracking is on
+
+      if (process.env.REACT_APP_HEAP_SITE_ID) {
```
1egoman: This is fine, though after taking a look at the `react-hotjar` source code, it looks like all it does is include the hotjar javascript snippet into the bundle, see https://github.com/abdalla/react-hotjar/blob/master/src/react-hotjar.js. Might it be a good idea to effectively do that ourselves too? Something like make a function above that's something like `initializeHotJar` that has the contents of the initialization code in it?

The argument for doing that is that we can remove a dependency, the cons are that we have to take on the complexity if something stops working (and then again, we may have to do this anyway if the package breaks as it doesn't look like its a super active project).

I'll let you make the decision on this.

PR #228, opened 2019-05-02
Den 7854 google calendar integration

cbigler:
Adds handling for Google Calendar room booking integration.


src/components/admin-integrations/index.tsx
```diff
@@ -62,6 +63,36 @@ export function AdminIntegrations({
     }
   }
 
+  function activateEditLink(item) {
```
1egoman: These functions should probably be moved outside of the component. Otherwise, when rendering, they will be instantiated every single time, which IMO in this situation is not needed.

In general, it's usually a better pattern to move those functions outside of the component and then add any values from the scope that it would otherwise inherit as parameters to the relocated function.

src/components/admin-integrations/index.tsx
```diff
@@ -62,6 +63,36 @@ export function AdminIntegrations({
...
+    }
+
+    if (item.name === "robin") {
+      return <ListViewClickableLink>Edit</ListViewClickableLink>;
+    }
+  }
+
+  function handleActivateEditClick(item) {
+    if (item.name == "teem") {
+      item.serviceAuthorization.id == null ? window.location.href = `https://app.teem.com/oauth/authorize/?client_id=${process.env.REACT_APP_TEEM_CLIENT_ID}&redirect_uri=${process.env.REACT_APP_TEEM_REDIRECT_URL}&response_type=code&scope=reservations` : null;
```
1egoman: This ternary probably should be made an if statement to be a bit more readable, something like:
```
if (item.name === 'teem') {
  window.location.href = `https://app.teem.com/oauth/authorize/?client_id=${process.env.REACT_APP_TEEM_CLIENT_ID}&redirect_uri=${process.env.REACT_APP_TEEM_REDIRECT_URL}&response_type=code&scope=reservations`
}
```

Also, `===` is preferred over `==`.

src/components/admin-integrations/index.tsx
```diff
@@ -62,6 +63,36 @@ export function AdminIntegrations({
...
+      if (item.serviceAuthorization.id == null) {
+        return doGoogleCalendarAuthRedirect();
+      }
+    } else {
+      onOpenModal(item.serviceAuthorization.id == null ? 'integrations-robin-create' : 'integrations-robin-update', {serviceAuthorization: item.serviceAuthorization, isDestroying: false})
+    }
+  }
+
+  function handleDeleteClick(item) {
+    if (item.name == "teem" || item.name == "google_calendar") {
```
1egoman: `===` is preferred over `==`.

src/components/admin-integrations/index.tsx
```diff
@@ -62,6 +63,36 @@ export function AdminIntegrations({
...
+
+    if (item.name === "robin") {
+      return <ListViewClickableLink>Edit</ListViewClickableLink>;
+    }
+  }
+
+  function handleActivateEditClick(item) {
+    if (item.name == "teem") {
+      item.serviceAuthorization.id == null ? window.location.href = `https://app.teem.com/oauth/authorize/?client_id=${process.env.REACT_APP_TEEM_CLIENT_ID}&redirect_uri=${process.env.REACT_APP_TEEM_REDIRECT_URL}&response_type=code&scope=reservations` : null;
+    } else if (item.name == "google_calendar") {
```
1egoman: Consider maybe making this condition `else if (item.name == "google_calendar" && item.serviceAuthorizationId === null) {` and avoiding the second if.

src/actions/integrations/google-calendar.ts
```diff
@@ -0,0 +1,6 @@
+import core from '../../client/core';
+
+
+export default function doGoogleCalendarAuthRedirect() {
+    core().get('/integrations/google_calendar/').then(resp => window.location.href = resp.data);
```
1egoman: Currently, there isn't anything special that will happen if this request fails. Is there any sort of special behavior that should happen? Like show an error message of some sort?

In addition, though it's a little pedantic and won't matter in this case, it's usually good form to return promises from a function so that any errors will bubble up to the caller.

cbigler:
@1egoman I've updated the PR with changes based on your feedback

PR #230, opened 2019-05-03
Space management time segments

1egoman:


src/actions/space-management/time-segments.ts
```diff
@@ -0,0 +1,142 @@
+import moment from 'moment';
```
1egoman: This file is where all the time segment update logic happens. It's called by the two new `space-management-update` and `space-management-create` actions and handles performing all the time segment and time segment group updates.

src/actions/space-management/time-segments.ts
```diff
@@ -0,0 +1,142 @@
...
+  // normalize start / end times on the next day onto the current day.
+  const secondsIntoDay = seconds % ONE_UTC_DAY_IN_SECONDS;
+
+  return moment.utc()
+    .startOf('day')
+    .add(seconds, 'seconds')
+    .format('HH:mm:ss');
+}
+
+export default function updateTimeSegments(id, operatingHoursLog) {
```
1egoman: The way this works is really wasteful but seems to be reliable. All actions are stored in a log, and then they are all replayed in order to get the server state matching the client's state. This means if someone moves the time segment slider around and lets go, and then moves it around again, both the first and second movement will be sent in that order.

src/actions/space-management/time-segments.ts
```diff
@@ -0,0 +1,142 @@
...
+          });
+
+          tempIdsToRealIds[entry.id] = timeSegmentGroupResponse.data.id;
+          return timeSegmentGroupResponse;
+
+        case TIME_SEGMENT_ASSIGN_TO_TIME_SEGMENT_GROUP:
+          const timeSegmentId = entry.timeSegmentId;
+          const timeSegmentGroupId = entry.timeSegmentGroupId;
+
+          const output = await Promise.all([
```
1egoman: This is the code that handles removing a time segment from all groups that is in prior to it being reassigned to the single group (ie, label) that it was given in the ui.

src/actions/space-management/time-segments.ts
```diff
@@ -0,0 +1,142 @@
...
+            })),
+
+            // Fetch the details for the time segment group
+            objectSnakeToCamel<DensityTimeSegmentGroup>(
+              (await core().get(`/time_segment_groups/${getId(timeSegmentGroupId)}`)).data
+            ),
+          ]);
+          const timeSegmentGroup = output[1];
+
+          return core().put(`/time_segment_groups/${getId(timeSegmentGroupId)}`, {
```
1egoman: "Append" the time segment to the time segment group. There's a [time of check to time of use](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) issue here that I think is unavoidable without serverside changes.

src/components/admin-locations-edit/index.tsx
```diff
@@ -27,15 +25,55 @@ import {
...
   Icons,
 } from '@density/ui';
 
+import spaceManagementUpdate from '../../actions/space-management/update';
+import { OperatingHoursItem, OperatingHoursLabelItem } from '../../actions/space-management/time-segments';
+
+export function calculateOperatingHoursFromSpace(
+  space: DensitySpace,
+  timeSegmentGroups: Array<DensityTimeSegmentGroup>,
+): Array<OperatingHoursItem> {
```
1egoman: This function is used to "deserialize" time segments into operating hours. These operating hours data structures are very similar to time segments but have a few fields that are easier to handle in the ui (like the times being in seconds, and the time segment group the time segment is assigned to as part of the data structure instead of requiring a lookup, etc)

src/components/admin-locations-edit/index.tsx
```diff
@@ -27,15 +25,55 @@ import {
...
+      daysAffected: tsm.days,
+      actionToPerform: null,
+    };
+  });
+}
+
 // Given a space and the currently logged in user, return the initial state of eitehr the edit or
 // new form.
-export function calculateInitialFormState(space, user): AdminLocationsFormState {
+export function calculateInitialFormState(space, user, timeSegmentGroups): AdminLocationsFormState {
```
1egoman: When the initial form state is calculated, calculate operating hours (using the function just above), operating hours labels, and initialize the log of operations that will be later used when updating the server.

src/components/admin-locations-new/index.tsx
```diff
@@ -64,10 +68,35 @@ const ALLOWED_SUB_SPACE_TYPES = {
 class AdminLocationsNewUnconnected extends Component<AdminLocationsNewProps, AdminLocationsFormState> {
   constructor(props) {
     super(props);
-    this.state = calculateInitialFormState({
-      parentId: props.newSpaceParent ? props.newSpaceParent.id : null,
-      spaceType: props.newSpaceType,
-    }, props.user);
+
```
1egoman: The new page now needs the same logic as in the edit page concerning if resources have loaded, since in order to calculate the initial form state we now need a list of all time segment groups.

src/helpers/space-hierarchy-formatter-deprecated/index.ts
```diff
@@ -0,0 +1,96 @@
+const SPACE_TYPE_SORT_ORDER = [
```
1egoman: Did a bunch of work cleaning up the hierarchy stuff. Now there is a separate set of helpers for what I'm calling the "deprecated" hierarchy formatter and the regular hierarchy formatter. I also modified the new one so that we no longer have to pass it the spaces collection when it runs, as it seemed that wouldn't be too hard.

src/actions/route-transition/admin-locations-new.ts
```diff
@@ -1,41 +1,62 @@
...
 export const ROUTE_TRANSITION_ADMIN_LOCATIONS_NEW = 'ROUTE_TRANSITION_ADMIN_LOCATIONS_NEW';
 
-export default function routeTransitionAdminLocationsNew(parentSpaceId, newSpaceType) {
-  return async (dispatch, getState) => {
-    const space = parentSpaceId && getState().spaces.data.find(space => space.id === parentSpaceId);
+async function getHierarchy() {
+  return (await core().get('/spaces/hierarchy/')).data.map(objectSnakeToCamel);
+}
+
+async function getSpaces() {
```
guscost: Aaahhh, duplicate things! Data access layer!

src/components/admin-locations-detail-modules-operating-hours-slider/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+      this.pressedButton = event.target.id.indexOf('start') >= 0 ? 'start' : 'end';
+
+      const trackBbox = track.getBoundingClientRect();
+      this.trackWidthInPx = trackBbox.width;
+      this.trackLeftPositionInPx = trackBbox.left;
+
+      // Add or subtract a small offset from the track left position. This effectively ensures that
+      // further handle movments are relative to the original cursor position so that the handle
+      // doesn't "jump" to the original cursor position when it is first moved.
+      const handleBbox = event.target.getBoundingClientRect();
```
guscost: This component is full of voodoo and is a bit scary, but at least it's encapsulated (except maybe for the use of the `id` attribute)/

src/components/admin-locations-detail-modules/index.tsx
```diff
@@ -53,9 +66,15 @@ export default function AdminLocationsDetailModule({title, error=false, actions=
...
       </div>
-      <div className={styles.moduleBody}>
-        {children}
-      </div>
+      {children}
+    </div>
+  );
+}
+
+export function AdminLocationsDetailModuleBody({includePadding=true, children}) {
```
guscost: I'm somewhat ambivalent about this abstraction vs. duplicating the card layout stuff in each individual module, but would like to break out all of the other specific modules into their own files like `operating-hours.tsx` is already, since this file is >700 lines now

src/components/admin-locations-detail-modules/index.tsx
```diff
@@ -1,15 +1,18 @@
...
 import { connect } from 'react-redux';
 import styles from './styles.module.scss';
 import FormLabel from '../form-label/index';
 import classnames from 'classnames';
-import TIME_ZONE_CHOICES from '../../helpers/time-zone-choices/index';
-import generateResetTimeChoices from '../../helpers/generate-reset-time-choices/index';
+
 import { UNIT_NAMES, SQUARE_FEET, SQUARE_METERS } from '../../helpers/convert-unit/index';
 
+import spaceHierarchyFormatter from '../../helpers/space-hierarchy-formatter/index';
```
guscost: Lots of leftover imports here which are not used anymore.

src/components/admin-locations-new/index.tsx
```diff
@@ -63,26 +60,19 @@ const ALLOWED_SUB_SPACE_TYPES = {
...
-  constructor(props) {
-    super(props);
-    this.state = calculateInitialFormState({
-      parentId: props.newSpaceParent ? props.newSpaceParent.id : null,
-      spaceType: props.newSpaceType,
-    }, props.user);
-  }
-
-  onChangeField = (key, value) => {
-    this.setState(s => ({...s, [key]: value}));
```
guscost: We could just use hooks™️ and not have to deal with this weird fake-class-but-halfway-functional nonsense.

src/components/day-of-week-selector/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  'Monday',
+  'Tuesday',
+  'Wednesday',
+  'Thursday',
+  'Friday',
+  'Saturday',
+  'Sunday',
+];
+
+export default function DayOfWeekSelector({ daysOfWeek, disabled=false, onChange }) {
```
guscost: Cool 👍

src/components/admin-locations-edit/index.tsx
```diff
@@ -267,21 +191,28 @@ class AdminLocationsEdit extends Component<AdminLocationsEditProps, AdminLocatio
...
-    async onSave(spaceFieldUpdate, spaceId) {
-      const ok = await dispatch(collectionSpacesUpdate(spaceFieldUpdate));
-      if (ok) {
-        dispatch(showToast({ text: 'Space updated successfully' }));
-      } else {
+    async onSave(spaceId, spaceFieldUpdate) {
+      const ok = await dispatch(collectionSpacesUpdate({
+        ...spaceFieldUpdate,
+        id: spaceId,
+        timeSegmentGroups: [],
```
guscost: Is this correct, or should it be something with labels?

src/components/admin-locations-new/index.tsx
guscost: Nevermind, that probably wouldn't help

src/components/admin-locations-detail-modules-operating-hours-slider/index.tsx
```diff
@@ -0,0 +1,233 @@
+import React, { Component, Fragment } from 'react';
+import classnames from 'classnames';
+import styles from './styles.module.scss';
+import moment from 'moment';
+import { Icons } from '@density/ui';
+
+import generateResetTimeChoices from '../../helpers/generate-reset-time-choices/index';
+
+const UTC_DAY_LENGTH_IN_SECONDS = moment.duration('24:00:00').as('seconds');
```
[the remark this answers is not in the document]
1egoman: Yes!

src/components/admin-locations-edit/index.tsx
1egoman: That is vestigial

src/components/admin-locations-detail-modules-operating-hours-slider/index.tsx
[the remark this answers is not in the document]
1egoman: Actually, maybe we should just put this in `space-time-utilities`? Either that or making a separate place sounds fine to me.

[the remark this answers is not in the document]
guscost: i'm thinking of starting with a catch-all `constants` helper that just exports a bunch of constants (with unambiguous names) for any possible thing, and we can move them from there into anywhere that makes more sense

src/components/admin-locations-edit/index.tsx
guscost: I like "vestigial"

PR #234, opened 2019-05-08
Space image upload

guscost:
Adds interactions for uploading space images.

https://github.com/DensityCo/dashboard/pull/234/files#diff-a930abaaec8f2b32eea55ba2045332ee contains the helper which requests an S3 upload link, uploads to S3, and then polls for the new media to show up as completed. A lambda function is responsible for processing the uploaded images (resizing to 800px) and POSTing back to complete the media upload, hence the polling until this is done.


src/helpers/media-files/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.readAsDataURL(file);
+    reader.onload = () => { resolve(reader.result as string); }; 
+  });
+}
+
+// Upload media using a signed S3 upload link, and poll until done
+export default async function uploadMedia(path: string, file: File | Blob, quantity=1, retries=10) {
+  const { upload_id, upload_url } = (await core().post(path)).data;
```
1egoman: Can you use `objectSnakeToCamel` here?

src/helpers/media-files/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  await core().put(upload_url, file, {headers: {'Content-Type': file.type}});
+
+  // Poll until at least one upload is done
+  let status: any = null;
+  for (let i = 0; i < retries; i++) {
+    await sleep(1000);
+    status = (await core().get(`/upload/${upload_id}`)).data;
```
1egoman: Can you use `objectSnakeToCamel` here?

src/components/form-label/index.tsx
```diff
@@ -16,7 +16,7 @@ export default function FormLabel({
   return (
     <div className={classnames(styles.formLabel, editable ? styles.formLabelEditable: null, className)}>
       <label className={styles.formLabelLabel} htmlFor={htmlFor}>
-        <span className={styles.formLabelLabelText}>{label}</span>
+        {typeof label === 'string' ? <span className={styles.formLabelLabelText}>{label}</span> : label}
```
1egoman: Curious as to why this was needed?

src/components/admin-locations-image-upload/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  // More info: https://stackoverflow.com/questions/32414308/updating-state-on-props-change-in-react-form
+  componentWillReceiveProps(nextProps) {
+    const newComponentView = this.calculateViewGivenProps.call(this, nextProps);
+
+    // Update if state changed.
+    if (newComponentView !== this.state.view) {
+      this.setState({view: newComponentView});
+    }
+  }
+  calculateViewGivenProps(props) {
```
1egoman: If you make this function an instance variable (I think that's the right word?) like the below, the above `.call(this, props)` is not needed and you can just invoke it normally (the `this` value will always be set to the `this` of the class instance)
```
calculateViewGivenProps = (props) => {
  // code here
}
```

src/components/form-label/index.tsx
guscost: The "Edit" link is right-aligned in the uploaded image preview, but I wanted to reuse this component/form convention.

src/components/admin-locations-image-upload/index.tsx
guscost: Fake classes and now fake class instances, bleh

guscost: I don't want to do this since it feels like I'm participating in something wrong. Feel free to refactor if you like these instance methods.

1egoman: Doesn't matter to me all that much, both are valid ways to solve the problem.

src/actions/collection/spaces/create.ts
```diff
@@ -30,6 +31,12 @@ export default function collectionSpacesCreate(item) {
         time_zone: item.timeZone,
         daily_reset: item.dailyReset,
       });
+      if (item.newImageFile) {
+        const upload = await uploadMedia(`/uploads/space_image/${response.data.id}`, item.newImageFile);
+        response.data.image = upload.media[0].signed_url;
```
1egoman: I think this needs to be `signedUrl` now that uploadMedia uses `objectSnakeToCamel`?

guscost: Nice catch, thanks

src/components/admin-locations-image-upload/index.tsx
1egoman: Doesn't matter to me all that much, both are valid ways to solve the problem.

guscost: Down with classes! Use hooks!

PR #237, opened 2019-05-16
Space management tags

1egoman:
Adds a "tags" module to the space management detail pages for each space. This allows users to enter tags they'd like to assign to a space.


src/actions/collection/spaces/update.ts
```diff
@@ -48,6 +49,17 @@ export default function collectionSpacesUpdate(item) {
         dispatch(hideToast(id));
       }
 
+      if (item.newTags) {
+        await Promise.all(item.newTags.map(async tag => {
```
1egoman: Tags cannot be updated, only created (assigned to a space) or deleted (unassigned from a space).

src/components/admin-locations-detail-modules/index.tsx
```diff
@@ -302,6 +305,65 @@ export function AdminLocationsDetailModulesGeneralInfo({spaceType, formState, on
...
+  onChangeTags,
+}) {
+  return (
+    <AdminLocationsDetailModule title={title}>
+      <TagInput
+        choices={tagsCollection.data.map(f => ({id: f.name, label: f.name}))}
+        tags={tags.filter(f => f.operationToPerform !== 'DELETE').map(f => ({id: f.name, label: f.name}))}
+        placeholder={placeholder}
+        emptyTagsPlaceholder={emptyTagsPlaceholder}
+        onCreateNewTag={name => {
```
1egoman: This logic below handles attaching the proper things to do to each tag as it is created. It's decently complicated, but at least there are only three possible states of `operatingToPerform` (`null` to do nothing, `CREATE` to create a tag / associate it with a space, `DELETE` to disassociate a tag with a space, and if it's the last to be deleted, the server deletes the tag outright)

src/reducers/tags/index.ts
```diff
@@ -0,0 +1,32 @@
+import objectSnakeToCamel from '../../helpers/object-snake-to-camel/index';
```
1egoman: Added a tags collection. I'm thinking that maybe it may make more sense for this to go into the space management reducer once we have the operating hours branch merged in, but that's a longer term discussion.

PR #243, opened 2019-05-29
Show still syncing notice and check for the service spaces every 5 se…

beweinreich:
Show "still syncing" notice and check for the service spaces every 5 seconds. Once the service spaces show up, and `lastSync !== null` we can allow the customer to wrap up the space mappings.


src/components/admin-space-mappings/index.tsx
```diff
@@ -32,112 +33,137 @@ import showModal from '../../actions/modal/show';
...
-  const serviceSpaceForService = (spaceMappings, service) => {
-    if (service) {
-      const spaceMappingsForService = spaceMappings.filter(spm => spm.serviceId == service.id);
-      if (spaceMappingsForService) {
-        return spaceMappingsForService[0];
+
+export class AdminSpaceMappings extends React.Component<any, any> {
+  
+  componentDidMount() {
+    var that: any = this;
```
guscost: 😱 you can use an arrow function as the argument to `setInterval` instead of aliasing `this` (arrow functions use the containing scope's `this`).

src/components/admin-space-mappings/index.tsx
```diff
@@ -32,112 +33,137 @@ import showModal from '../../actions/modal/show';
...
+    setInterval(function() {
+      if (that.props.currentService.serviceAuthorization.lastSync === null) {
+        that.props.onCheckForIntegrationLastSync();
       }
-    }
-    return null;
+    }, 5000)
```
guscost: Semicolon plz

src/components/admin-space-mappings/index.tsx
```diff
@@ -32,112 +33,137 @@ import showModal from '../../actions/modal/show';
...
+        <div className={styles.adminIntegrationsList}>
+          <div className={styles.adminIntegrationsSectionHeader}>Link your room-booking spaces</div>
+          
+          {currentService && currentService.serviceAuthorization.lastSync === null ? (
+            <div className={styles.centeredMessage}>
+              <div className={styles.integrationNotice}>
+                We are syncing your {currentService.displayName} spaces with Density. This page will automatically refresh when the sync is complete.
+              </div>
+            </div>
+          ) : null}
```
guscost: Why not show a toast (there's a `showToast()` helper) for this message?

beweinreich: I want the message to persist and then disappear once it's done. A toast that shows up and disappears doesn't have the same effect.

A toast that persists would just bug me :)

src/components/admin-space-mappings/index.tsx
```diff
@@ -32,112 +33,137 @@ import showModal from '../../actions/modal/show';
...
-    if (service) {
-      const spaceMappingsForService = spaceMappings.filter(spm => spm.serviceId == service.id);
-      if (spaceMappingsForService) {
-        return spaceMappingsForService[0];
+
+export class AdminSpaceMappings extends React.Component<any, any> {
+  
+  componentDidMount() {
+    var that: any = this;
+    setInterval(function() {
```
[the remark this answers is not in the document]
beweinreich: Got it. Will update.

src/components/admin-space-mappings/index.tsx
guscost: You can `hideToast()` programmatically if you want, let me find an example.

PR #244, opened 2019-05-29
Resizing based on ref and window listeners

beweinreich:
- Made both pages class based. 
- Wrapped them in a div with a ref (since you cant ref a functional component)
- Attached window listeners to change `onResize`
- Change width of sidebar

### DIFF
--- a/src/components/dashboard/index.tsx
+++ b/src/components/dashboard/index.tsx
@@ -195,137 +195,167 @@ function DashboardMainScrollViewContent({
   }
 }
 
-export function Dashboard({
-  dashboards,
-  selectedDashboard,
-  activeModal,
+export class Dashboard extends React.Component<any, any> {
+  private pageContainerRef: React.RefObject<HTMLInputElement>;
 
-  date,
-  sidebarVisible,
-  resizeCounter,
-  settings,
-  isDemoUser,
+  constructor(props) {
+    super(props);
+    this.pageContainerRef = React.createRef();
+    this.state = {
+      pageSize: 0,
+    };
+  }
 
-  onDashboardChangeWeek,
-  onChangeSidebarVisibility,
-  onCloseModal,
-  onShowModal,
-}) {
+  componentDidMount() {
+    window.addEventListener('resize', this.onResize.bind(this));
+  }
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.onResize.bind(this));
+  }
 
-  const rootDiv: any = document.getElementById("root");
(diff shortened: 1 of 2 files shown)


src/components/dashboard/index.tsx
```diff
@@ -195,137 +195,167 @@ function DashboardMainScrollViewContent({
...
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.onResize.bind(this));
+  }
 
-  onDashboardChangeWeek,
-  onChangeSidebarVisibility,
-  onCloseModal,
-  onShowModal,
-}) {
+  onResize() {
```
1egoman: If you make this function an attribute on the class with `onResize = () => {}` you don't need the `.bind(this)`s when you call it!

src/components/dashboard/index.tsx
```diff
@@ -195,137 +195,167 @@ function DashboardMainScrollViewContent({
...
 }
 
-export function Dashboard({
-  dashboards,
-  selectedDashboard,
-  activeModal,
+export class Dashboard extends React.Component<any, any> {
+  private pageContainerRef: React.RefObject<HTMLInputElement>;
+
+  constructor(props) {
```
1egoman: Kinda nitpicky and not all that important, but you can define these attributes inine and don't really need to define a constructor here. Doing it this way also means that you don't need to define the explicit type on `pageContainerRef`.

```
pageContainerRef = React.createRef();
state = {
  pageSize: 0,
};
```

Not all that important though, what you have works fine and is just as easy to understand!

src/components/explore/index.tsx
```diff
@@ -147,93 +147,124 @@ function pruneHierarchy(spaceTree, matchedSpaceIds) {
...
-  let filteredSpaces = spaceHierarchy.data;
-  if (spaces.filters.search) {
-    const matchedSpaceIds = fuzzy.filter(
-      spaces.filters.search,
-      spaces.data,
-      { pre: '<', post: '>', extract: x => x['name'] }
-    ).map(x => x.original['id']);
-    const filteredSpacesCopy = JSON.parse(JSON.stringify(filteredSpaces));
-    filteredSpaces = filteredSpacesCopy.map(x => pruneHierarchy(x, matchedSpaceIds)).filter(x => x);
+export class Explore extends React.Component<any, any> {
```
1egoman: Same notes on this component as the other one

src/components/dashboard/index.tsx
beweinreich: Makes sense

PR #250, opened 2019-06-05
grammar: capitalize with sentence-case

guscost:


src/actions/collection/digest-schedules/create.ts
```diff
@@ -32,7 +32,7 @@ export default function collectionDispatchSchedulesCreate({
         time_zone: timeZone,
       });
 
-      mixpanelTrack('Email Digest Created', {
```
1egoman: Check with brian before changing this one, I bet this will cause issues with analytics showing up in a different place in mixpanel

guscost: not even going there, thanks for the heads up

PR #251, opened 2019-06-07
code: fetch all objects prototype

guscost:


guscost:
Moving into a single PR with the button stuff

PR #253, opened 2019-06-07
New buttons and fetchObject(s) helper

guscost:
Updates Density UI to 17.0.0, and migrates it to use the changed Button component in that release. Also updates dashboard to use a new fetchAllObjects helper with caching.

PR #254, opened 2019-06-11
Explore page date decreasing

1egoman:
Attempts to fix an issue where the trends or daily page's selected date or date range will change when a space in a different time zone is selected.

*NOTE*: Do not merge! This is quite broken.

Also, This PR has a dependency on https://github.com/DensityCo/dashboard/pull/253 being merged first.


guscost:
Apparently this one broke the whole explore page, we'll have to try again w/ all the new changes to it.

PR #260, opened 2019-06-19
alert management (and new accounts page)

guscost:
Also includes #257


src/actions/collection/alerts/create.ts
```diff
@@ -0,0 +1,56 @@
...
+  isOneShot,
+  cooldown,
+  meta
+}: DensityNotification) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_ALERTS_CREATE });
+
+    let alert, errorThrown;
+    try {
+      alert = await core().post(getNotificationsURL(), {
```
1egoman: BTW, this can use v2 now because of https://github.com/DensityCo/core-api/pull/935

src/actions/collection/alerts/update.ts
```diff
@@ -0,0 +1,49 @@
...
+  isOneShot,
+  cooldown,
+  meta
+}: DensityNotification) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_ALERTS_UPDATE });
+
+    let alert, errorThrown;
+    try {
+      alert = await core().put(`${getNotificationsURL()}/${id}`, {
```
1egoman: Can use `v2` here too: https://github.com/DensityCo/core-api/pull/935

src/actions/collection/alerts/load.ts
```diff
@@ -0,0 +1,33 @@
...
+  return `${baseV1}/notifications`;
+}
+
+export default function collectionAlertsLoad() {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_ALERTS_LOAD });
+
+    let alerts, errorThrown;
+    try {
+      alerts = await fetchAllObjects<DensityNotification>(getNotificationsURL());
```
1egoman: Can use `v2` here too: https://github.com/DensityCo/core-api/pull/935

src/actions/collection/alerts/destroy.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const COLLECTION_ALERTS_DESTROY = 'COLLECTION_ALERTS_DESTROY';
+
+export default function collectionAlertsDestroy(alert) {
+  return async dispatch => {
+    dispatch({ type: COLLECTION_ALERTS_DESTROY });
+
+    let errorThrown;
+    try {
+      await core().delete(`${getNotificationsURL()}/${alert.id}`);
```
1egoman: Can use `v2` here too: https://github.com/DensityCo/core-api/pull/935

src/components/account/password-form.bak
```diff
@@ -0,0 +1,10 @@
+{this.props.activeModal.name === 'account-password-reset' && this.props.activeModal.visible ? <div className={styles.accountPasswordResetToast}>
```
1egoman: Can this file be deleted?

src/components/explore-alert-popup-list/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { DensitySpace } from '../../types';
+
+type ExploreAlertPopupListProps = {
+  alerts: {
+    view: 'VISIBLE' | 'LOADING' | 'ERROR',
+    data: Array<any>,
+    error: string | null,
+  },
+  selectedSpace: DensitySpace,
+  onEditAlert: any,
```
1egoman: Kind of a nitpick, but I've started typing callbacks like this by having them return `void` (which as far as I can tell, is the type for "a function that can return anything and the return value shouldn't be used")

src/components/account/password-form.bak
guscost: Whoops, yep.

src/components/explore-alert-popup-list/index.tsx
guscost: tricky part is when they accept a react pseudo-event or a value or whatever, but those are typeable too

PR #264, opened 2019-06-24
code: work on optimizing components

guscost:
Optimizes components, mostly on the Explore page, by removing unnecessary state and using `React.memo` on pure functional components.

To achieve some stateless pages/components that need to measure their own width, there is also a new `autoWidthHoc` helper component, which takes the inner component to render (which will be passed a `width` prop automatically) as well as a timeout in milliseconds (the HOC will re-render after debouncing by this many ms).


src/components/explore-space-detail-utilization-card/index.tsx
```diff
@@ -48,7 +49,7 @@ export const LOADING = 'LOADING',
              REQUIRES_CAPACITY = 'REQUIRES_CAPACITY',
              ERROR = 'ERROR';
 
-export class ExploreSpaceDetailUtilizationCard extends React.Component<any, any> {
+export class ExploreSpaceDetailUtilizationCard extends React.PureComponent<any, any> {
```
de1ux: dont think you need the second `any` type parameter

de1ux:
while I don't have the context for some of these changes, I do 💯the `Memo` and `PureComponent` usage

guscost:
Got verbal 👍 , merging and testing in staging

PR #265, opened 2019-06-25
Added in org ID and name to intercom data

beweinreich:
We will be syncing org ID across Intercom, Mixpanel, and Salesforce. This pushes org ID and name to intercom fields.

### DIFF
--- a/src/components/intercom/index.tsx
+++ b/src/components/intercom/index.tsx
@@ -10,8 +10,8 @@ export function IntercomDensity({ user }) {
       user_id: user.data.id,
       email: user.data.email,
       name: user.data.fullName,
-      org_id: user.data.organization.id,
-      org_name: user.data.organization.name,
+      organization_id: user.data.organization.id,
+      organization_name: user.data.organization.name,
     }
   }
   if (process.env.REACT_APP_INTERCOM_APP_ID) {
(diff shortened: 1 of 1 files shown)


src/components/intercom/index.tsx
```diff
@@ -9,7 +9,9 @@ export function IntercomDensity({ user }) {
     intercomUser = {
       user_id: user.data.id,
       email: user.data.email,
-      name: user.data.fullName
+      name: user.data.fullName,
+      org_id: user.data.organization.id,
```
guscost: Could we use `organization_name` as it's the convention in most other places, or does this have to match a convention among the user analytics tools?

beweinreich: yeah, its already org_name in the other tools, so I'm just being consistent.

beweinreich: sorry— to clarify, it's already org_name in Mixpanel and Intercom, so I want to keep it consistent. 

yeah yeah, i know im usually the big bad wolf when it comes to abbreviations in this stuff, but a lot of these fields in these third party systems have max key sizes :(

beweinreich: waaaiiiit. i lied! mixpanel actually has `organization_id` and `organization`. ugh, i wish I had done `organization_name`. Just checked MP and SFDC and they look like they can handle that length. 

okay, changing!

PR #266, opened 2019-06-25
add "space parent" picker

guscost:


1egoman:
Rob says that he's prefer this to be a select box, but it's likely we will merge this anyway and then will fix it up at a later date.

guscost:
There are mockups?

guscost:
Aha, so there are

src/components/admin-locations-detail-modules/styles.module.scss
```diff
@@ -8,6 +8,8 @@
   width: 818px;
   border: 1px solid $grayLight;
   border-radius: $borderRadiusBase;
+
+  // overflow: hidden;
```
1egoman: Can this be removed?

PR #270, opened 2019-07-05
Updating alert verbiage

beweinreich:
@guscost I'm looking to simplify this alert a bit. I'd like for the default to be "one-time." I changed the verbiage as well to make it clear that the third field is about how many times we'll notify the user.

Completes: https://densityio.atlassian.net/browse/DEN-8436

I'm having trouble making “Notify me at most” perform the following logic:
- Selecting "Once" changes it to one-time + null cooldown.
- Selecting anything else changes it to "one-time: false" and adds the cooldown.

The other issue I'd like to tackle here:
- https://densityio.atlassian.net/browse/DEN-8444

### DIFF
--- a/src/components/explore-alert-management-modal/index.tsx
+++ b/src/components/explore-alert-management-modal/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 
 import {
   AppBar,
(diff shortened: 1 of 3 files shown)


src/components/explore-alert-management-modal/index.tsx
```diff
@@ -119,46 +116,47 @@ export function ExploreAlertManagementModalRaw({
...
-                    'escalationDelta',
-                    e.target.value.replace(/[^0-9]/, '')
-                  )}
-                />
-                <div style={{width: 8}}></div>
-                {parseInt(alert.meta.escalationDelta) === 1 ? 'person' : 'people'}
-              </div>}
-            />
-          </div> : null
+        {alert.triggerType === GREATER_THAN && parseInt(alert.cooldown) !== -1 ?
```
1egoman: Can you make this parseInt be `parseInt(alert.cooldown, 10)`?

PR #277, opened 2019-07-17
Density UI ListView

guscost:
Port the ListView to use the new Density UI control (which renders an actual <table> element). Tested in all supported browsers.


src/components/admin-locations-root-detail/index.tsx
```diff
@@ -1,20 +1,23 @@
...
+import { Icons, ListView, ListViewColumn } from '@density/ui';
 
 
 function SpaceList({ user, spaces, renderedSpaces }) {
   return (
     <div className={styles.spaceList}>
-      <ListView data={renderedSpaces}>
+      <ListView
+        data={renderedSpaces}
+        onClickRow={item => window.location.href = `#/admin/locations/${item.id}`}
```
de1ux: this is cool, now I don't have to set a clickable event for each column!

de1ux:
This was helpful to see how the ListView is intended to be used -- I'll make some adjustments in Sensei to follow suite

guscost:
@de1ux suggestions are welcome too! I think this is approaching "stable" status but we can always refactor to add features or break things apart as it evolves.

guscost:
Also it needs real documentation, trying to fit that in this week...

PR #294, opened 2019-08-08
Dpu Status

beweinreich:
Added in WiFi and LAN status to the DPU status page

Screen Shot 2019-08-08 at 1 31 10 PM

### DIFF
--- a/src/components/admin-device-status/index.tsx
+++ b/src/components/admin-device-status/index.tsx
@@ -19,6 +19,24 @@ function getStatusColor(status) {
   }
 }
 
+function renderNetworkAddressElement(item, networkInterface) {
+  if (item.networkAddresses) {
+    let networks = item.networkAddresses.filter(na => na.if === networkInterface);
+    if (networks.length > 0) {
+      return (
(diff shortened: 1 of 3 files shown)


src/components/admin-device-status/index.tsx
```diff
@@ -19,6 +19,24 @@ function getStatusColor(status) {
   }
 }
 
+function renderNetworkAddressElement(item, networkInterface) {
```
guscost: This could be a pure functional component with two props.

src/components/admin-device-status/index.tsx
```diff
@@ -62,6 +80,16 @@ export function AdminDeviceStatus({
           }).map(space => {
             return <div style={{marginRight: 10}} key={space.id}>{space.name}</div>
           })} />
+        <ListViewColumn
+          id="Ethernet"
+          width={280}
+          template={item => renderNetworkAddressElement(item, 'eth0')}
```
guscost: If this was a functional component, it could be `item => <NetworkAddressElement item={item} interface="eth0" />`

PR #312, opened 2019-08-13
new state management

guscost:
New state management using RxJS.

See https://github.com/DensityCo/dashboard/blob/feat-new-state-management/docs/RxState.md


docs/RxState.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Interop with Redux
+
+While we're still porting parts of the dashboard from Redux, all actions dispatched to Redux will also be emitted by the RxJS action stream. Conversely, all actions sent using the `useRxDispatch` hook will be sent to the Redux store.
+
+There are a few caveats around "thunk actions" with this interoperability:
+
+- First, thunk actions are not used in the new pattern and will be ignored by RxJS.
+- As a consequence, thunk actions dispatched using `useRxDispatch` will not return values to the caller even if the call is awaited.
+- Because of the order of Redux middleware, it's likely that thunk actions dispatched using `useRxDispatch` won't work at all...
```
guscost: We have talked about a departure from using thunk actions, and the lack of support here reflects that decision.

docs/RxState.md
```diff
@@ -0,0 +1,185 @@
...
+To only see new actions, use `localStorage.debug = 'rx:action,rx:state,rx:info';`
+
+## Interop with Redux
+
+While we're still porting parts of the dashboard from Redux, all actions dispatched to Redux will also be emitted by the RxJS action stream. Conversely, all actions sent using the `useRxDispatch` hook will be sent to the Redux store.
+
+There are a few caveats around "thunk actions" with this interoperability:
+
+- First, thunk actions are not used in the new pattern and will be ignored by RxJS.
+- As a consequence, thunk actions dispatched using `useRxDispatch` will not return values to the caller even if the call is awaited.
```
guscost: We have talked about a departure from thunk actions, and the lack of support here is because of that decision.

src/actions/collection/digest-schedules/update.ts
```diff
@@ -37,11 +38,11 @@ export default function collectionDispatchSchedulesUpdate({
...
       console.error(errorThrown);
-      dispatch(collectionDispatchSchedulesError(errorThrown));
-      return false;
+      dispatch(collectionDigestSchedulesError(errorThrown));
+      dispatch(showToast({ type: 'error', text: `Whoops! That didn't work.` }));
     } else {
-      dispatch(collectionDispatchSchedulesPush(schedule.data));
-      return true;
+      dispatch(collectionDigestSchedulesPush(schedule.data));
+      dispatch(showToast({ text: 'Digest saved.' }));
```
guscost: Because these thunk actions are dispatched with `useRxDispatch` (but processed by Redux), they cannot return success/failure flag to the caller.

src/components/admin-user-management-detail/index.tsx
```diff
@@ -26,82 +26,53 @@ import {
...
+}
 
-  render() {
-    const {
-      spaces,
-      spaceHierarchy,
-      user,
-      users,
-      selectedUser,
+export default function AdminUserManagementDetail() {
```
guscost: Converted to a functional component with hooks, mostly because this connects to both Redux and Rx stores.

src/components/spaces/index.tsx
```diff
@@ -74,28 +69,19 @@ function ExploreSpacePage({ activePage }) {
...
+  spaceReports,
+  selectedSpace,
+  activePage,
+  activeModal,
+  //resizeCounter,
+  onCloseModal,
+  onFilterSpaces,
+  onUpdateReportController,
+  onCalculateReportData,
+}) {
```
guscost: Converted this back to a connected component, just so that we won't have three patterns for connecting components (Redux connect HOC, Redux hooks, RxJS hooks) right out of the gate.

src/rx-actions/users/read-one.ts
```diff
@@ -0,0 +1,36 @@
...
+    }) as any);
+  } else {
+    dispatch({
+      type: UserActionTypes.USER_MANAGEMENT_USERS_PUSH,
+      user: objectSnakeToCamel<DensityUser>(response.data),
+    });
+    dispatch({
+      type: UserActionTypes.USER_MANAGEMENT_SET_EDITOR,
+      user: objectSnakeToCamel<DensityUser>(response.data),
+    });
```
guscost: This is a bit silly/sloppy, there's no use case for loading a single user in the dashboard _except_ when it's for that user's detail/edit page, so maybe this should set a "selectedUserId" piece of state or something rather than copying the same data to two places. Feel free to update this branch if you want to mess with it.

guscost:
@1egoman Agree with both of those. Let's be sure to keep an ongoing "type concerns" slot reserved in the frontend meeting, similar to the "custom UI context concerns".

For the second item, see `useRxObservable` https://github.com/DensityCo/dashboard/blob/6fbef248919516f9edfbba2761c753623040b3d1/docs/RxState.md#subscribeconnect for one way that store streams can be extended. There will definitely be more examples of these use cases, and while I'm probably going to fall on the side of moving carefully/avoiding new patterns more often than not, it's just because I want to keep things predictable as much as possible, and it doesn't mean you shouldn't find other use cases and drive them to adoption.

PR #327, opened 2019-08-19
code: explicit any generic type

guscost:


guscost:
This was verbally approved 👍

PR #329, opened 2019-08-20
Add handling for Condeco integration

cbigler:
Adds Condeco integration icon and add/edit/delete handling.

This integration has manual steps to configure, so it is intentional that they cannot activate it from the integration list (we're handing setup manually for out first Condeco customer [Shell]).


beweinreich:
... @cbigler whats the point of adding it to the Dashboard if you can't do anything with it? Just curious

src/components/admin-integrations/index.tsx
```diff
@@ -77,7 +74,7 @@ function handleActivateEditClick(item, onOpenModal) {
 }
 
 function handleDeleteClick(item, onOpenModal) {
-  if (item.name === "teem" || item.name === "google_calendar" || item.name === 'outlook') {
+  if (["teem", "google_calendar", "outlook", "condeco"].includes(item.name)) {
```
guscost: Single quotes I guess

PR #331, opened 2019-08-20
RxStore post-update stream

jshanley:
- changed order of arguments to `createRxStore` so that the name of the store is first
- created a class `RxStore` which extends `BehaviorSubject` and exposes a public member called ~`postUpdate$`~ `postUpdateStream` which is a stream of `{state, action}` objects that are published after each update


src/rx-stores/alerts.ts
```diff
@@ -9,7 +9,7 @@ const initialState: AlertState = {
   data: [],
 };
 
-export default createRxStore<AlertState>(initialState, (state, action) => {
+export default createRxStore<AlertState>('alertsStore', initialState, (state, action) => {
```
jshanley: @guscost potential mis-name here.. should it be `alertsStore` or `alertStore`... or something else?

src/rx-stores/index.ts
```diff
@@ -19,21 +19,39 @@ export const actions = new Subject<GlobalAction>();
...
+    super(initialValue);
+    this.postUpdate$ = new Subject<{ state: T, action: A }>()
+
+    action$.subscribe((action: A) => {
+      const nextState = reducer(this.value, action);
+      // Tricky convention here, store skips update if the reducer returns the special Symbol
+      if (nextState !== skipUpdate) {
+        storeLog(displayName, nextState);
+        this.next(nextState);
+        this.postUpdate$.next({
```
1egoman: One thing to note - I'm not sure if having a `.next` inside of subscribe is ideal here. This might be the more "rxjs-idiomatic" way to do this:

```
this.postUpdateStream = actionStream.pipe(
  scan(([lastAction, state], action) => [action, reducer(state, action)]),
  filter(([lastAction, state]) => state !== skipUpdate),
  tap(([lastAction, state]) => storeLog(displayName, state)),
  map(([action, state]) => ({ state, action })),
);
```

I'm not sure if this is a good idea necesarily, but effectively then you could get rid of the subject as the value that the subject "emits" is derived from the action stream anyway.

Maybe related, I'm still trying to understand all this but I think this is the sort of concern I have about having a `.next` call inside a `.subscribe` like this: https://medium.com/@benlesh/hot-vs-cold-observables-f8094ed53339

src/rx-stores/index.ts
```diff
@@ -19,21 +19,39 @@ export const actions = new Subject<GlobalAction>();
...
+// Not exporting this class, as the helper `createRxStore` is preferred
+class RxStore<T, A extends { type: string }> extends BehaviorSubject<T> {
+
+  public postUpdateStream: Subject<{ state: T, action: A }>
+
+  constructor(displayName: string, initialValue: T, reducer: (state: T, action: A) => T | typeof skipUpdate, action$: Subject<A>) {
+    super(initialValue);
+    this.postUpdateStream = new Subject<{ state: T, action: A }>()
+
+    action$.subscribe((action: A) => {
```
1egoman: Maybe this should also be `actions` or `actionStream` to match `postUpdateStream`?

src/rx-stores/alerts.ts
guscost: Eh, alertsStore maybe

src/rx-stores/index.ts
```diff
@@ -19,21 +19,39 @@ export const actions = new Subject<GlobalAction>();
 // Reducer functions can return this symbol to skip updating
 export const skipUpdate = Symbol('skipUpdate');
 
+
+// Not exporting this class, as the helper `createRxStore` is preferred
+class RxStore<T, A extends { type: string }> extends BehaviorSubject<T> {
```
guscost: Silly classes...

src/rx-stores/index.ts
guscost: Will scan work? Didn't we try this in planner and it seemed to not do what we all thought the spec said it should do?

1egoman: UPDATE: I skipped over `this.next(nextState);` when I reviewed this before, and my code example doesn't handle that. I was assuming that the action stream was being mapped directly into the post update stream and there's more than that going on here.

guscost:
We've moved away from using a subclass to expose this stream, in favor of piping the action stream and switchMap-ping to combine with whatever store(s) are needed. This works fine and is appropriately "native" RxJS (and therefore a bit esoteric) when it is the correct approach.

guscost:
On the other hand, we're making the `.value` accessor on the Store BehaviorSubject throw an error that prompts the user to try an `.imperativelyGetValue()` replacement, if they really do want to imperatively access the store. This will be a nice nudge away from using this pattern, but not close it off completely.

PR #332, opened 2019-08-22
Analytics control bar

1egoman:
Screen Shot 2019-08-22 at 11 52 34 AM

- Adds a new component called `<AnalyticsControlBar />` that renders a control containing a number of "filters" used to adjust an analytics query.
  - There are a number of subcomponents, including a space selector, a date range picker, and an interval selector.

- Added react storybook to this project, and made a few stories for this analytics control bar representing some edge cases and some of the subcomponents of it.


src/components/analytics-control-bar/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+          value={dateRange}
+          onChange={onChangeDateRange}
+        />
+        <AnalyticsIntervalSelector
+          value={interval}
+          onChange={onChangeInterval}
+        />
+      </div>
+      <div className={styles.analyticsControlBarSection}>
+        {/* FIXME: put actual icons and  buttons here, point this out in a review! */}
```
1egoman: I'm not sure if it makes sense to add the actual buttons here yet until the save functionality has been designed and we know what will be going here. What exists now is here only to mock up roughly what should be here from a size perspective.

src/components/analytics-control-bar-space-filter/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+import SpacePicker from '../space-picker';
+import Filter, { FilterBold } from '../analytics-control-bar-filter';
+import {
+  ItemList,
+  ARROW_TEMPLATE,
+  MultipleSelectItemList,
+  BackButton,
+  FilterDeleteButton,
+  SubmitButton,
+} from '../analytics-control-bar/utilities';
```
guscost: Seems contorted/wrong that this imports something from the `../analytics-control-bar/utilities` module, but then the index file in _that same_ folder imports this module.

package.json
```diff
@@ -14,6 +14,7 @@
     "@density/reports": "^8.1.1",
     "@density/ui": "^18.1.3",
     "@mapbox/mapbox-gl-geocoder": "^4.0.0",
+    "@storybook/addon-storysource": "^5.1.11",
```
guscost: Move to devDependencies? This probably belongs there.

src/components/analytics-control-bar-space-filter/index.tsx
1egoman: Yea I wasn't sure where to put that stuff. I could make a `analytics-control-bar-utilities` component maybe that doesn't have a default export?

1egoman: Or maybe another option could be to export all those utility components within `analytics-control-bar`'s `index.tsx`.

guscost: But that's still a circular import (not that it doesnt work, just same feeling about it).

What if it's just an `analytics-control-bar-elements` folder or something.

1egoman: Are you saying a directory like `src/components/analytics-control-bar-elements` that contains an `index.tsx` file containing all of the components within `src/components/analytics-control-bar/utilities.tsx` (sounds similar to my first proposal)?

PR #354, opened 2019-09-03
Analytics tabs

1egoman:
Adds a tab bar which will be used to display a list of analytics queries that are open.

### DIFF
--- a/src/components/analytics-tabs/story.tsx
+++ b/src/components/analytics-tabs/story.tsx
@@ -22,6 +22,28 @@ function State({ initialState, children }) {
   )
 }
 
+
+function handleCloseReport(state, id) {
+  const newReports = state.reports.filter(r => r.id !== id)
+  let newActiveId = state.activeId;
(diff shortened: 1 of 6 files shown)


src/components/analytics-tabs/story.tsx
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable import/first */
```
jshanley: what's this?

src/components/analytics-tabs/story.tsx
```diff
@@ -0,0 +1,102 @@
...
+import React, { Fragment, useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import uuid from 'uuid';
+
+import AnalyticsTabs from './index';
+
+function State({ initialState, children }) {
```
jshanley: yeah I was bummed that this kept getting re-defined. Once we get all merged up we can use the storybook helpers for this

PR #358, opened 2019-09-04
Initial Analytics Page

1egoman:
**NOTE: Before merging this, switch the base of this pull request to `trunk`. It's not `trunk` right now to make it easier to review!**

Adds an initial analytics page to the dashboard, visible at `#/analytics`:
- When someone visits the page, a route transition action is dispatched which the analytics store listens for. The store then uses two new helpers, `loadSpaces` and `loadSpaceHierarchy` to fetch their particular resources.
  - In effect, if we were to make these helpers used in other parts in the app, we could start to make a system where different pages would depend on different "resources", and when visited, each page would have a "list" of resources that it would require in order to show the page. This is sort of an "inversion of control" / "dependency injection" sort-of idea.
  - I think these helpers need better names though.
- In order to be able to dispatch these legacy spaces and space hierarchy actions, I needed to give them types. I added a `src/types/redux.ts` file where any redux actions that are dispatched can have their types defined and this resulting `ReduxAction` is included in the `GlobalAction` union.

- On the analytics page, I rendered both the control bar and tab bar, and put together a bit of temporary logic just so I could see how they worked together. In the process of doing that, I fixed a few cross-browser bugs in the control bar.


src/rx-stores/index.ts
```diff
@@ -28,6 +23,16 @@ export class StoreSubject<S> extends BehaviorSubject<S> {
...
 
+// LEGACY: "define rules that ... enforce a strict interface between existing
+// and modern code so it [is] easy to understand their relationship"
+// From https://slack.engineering/rebuilding-slack-on-the-desktop-308d6fe94ae4
+let reduxStore;
+
+// LEGACY: get the redux store as an observable that emits state updates
+let rxStoreSubscriptionUnsubscribe;
+export type ReduxState = Any<FixInRefactor>;
+export let RxReduxStore = new StoreSubject<ReduxState>({});
```
1egoman: I added a version of the redux store as an observable that can be used just like any of the other stores in a component.

src/rx-stores/index.ts
```diff
@@ -28,6 +23,16 @@ export class StoreSubject<S> extends BehaviorSubject<S> {
...
 }
 
+// LEGACY: "define rules that ... enforce a strict interface between existing
+// and modern code so it [is] easy to understand their relationship"
+// From https://slack.engineering/rebuilding-slack-on-the-desktop-308d6fe94ae4
+let reduxStore;
+
+// LEGACY: get the redux store as an observable that emits state updates
+let rxStoreSubscriptionUnsubscribe;
+export type ReduxState = Any<FixInRefactor>;
```
1egoman: Not sure where this should come from...

src/types/redux.ts
```diff
@@ -0,0 +1,16 @@
+import { DensitySpace } from '.';
+// import { COLLECTION_SPACES_SET } from '../actions/collection/spaces/set';
+// import { COLLECTION_SPACES_ERROR } from '../actions/collection/spaces/error';
+
+// FIXME: The below 'COLLECTION_SPACES_SET' and 'COLLECTION_SPACES_ERROR' should use the constants
+// imported above, but if I do that, it types it as `string` and not the constant string it is set
+// to. Please point this point in a review!
+export type CollectionSpacesSet = { type: 'COLLECTION_SPACES_SET', data: Array<DensitySpace> };
```
1egoman: I need some help with this, see the `FIXME` comment above.

src/rx-stores/analytics.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const AnalyticsStore = createRxStore<AnalyticsState>('AnalyticsStore', {}, (state, action) => skipUpdate);
+export default AnalyticsStore;
+
+const routeTransitionStream = actions.pipe(
+  filter(action => action.type === AnalyticsActionType.ROUTE_TRANSITION_ANALYTICS),
+  switchMap(() => RxReduxStore.pipe(take(1))),
+);
+
+const loadSpaces = () => (source: Observable<ReduxState>) => source.pipe(
```
1egoman: I think if we decide these `loadSpaces` / `loadSpaceHierarchy` helpers are good we should move them into another file, but I'd like to keep them here for now.

src/rx-stores/analytics.ts
```diff
@@ -0,0 +1,58 @@
...
+import createRxStore, { rxDispatch, actions, skipUpdate, RxReduxStore, ReduxState } from './index';
+import { DensitySpace, DensitySpaceHierarchyItem } from '../types';
+import { AnalyticsState, AnalyticsActionType } from '../types/analytics';
+import fetchAllObjects from '../helpers/fetch-all-objects';
+
+import collectionSpacesError from '../actions/collection/spaces/error';
+import collectionSpacesSet from '../actions/collection/spaces/set';
+import collectionSpaceHierarchySet from '../actions/collection/space-hierarchy/set';
+
+const AnalyticsStore = createRxStore<AnalyticsState>('AnalyticsStore', {}, (state, action) => skipUpdate);
```
1egoman: The analytics store handles no actions at the moment.

src/rx-stores/analytics.ts
```diff
@@ -1,15 +1,58 @@
...
-  // FIXME: Add route transition logic in here
-  console.log('ROUTE TRANSITION LOGIC HERE');
-});
+  switchMap(() => RxReduxStore.pipe(take(1))),
+);
+
+const loadSpaces = () => (source: Observable<ReduxState>) => source.pipe(
+  switchMap(reduxState => {
+    if (reduxState.spaces.view !== 'VISIBLE' && reduxState.spaces.data.length > 1) {
+      return of();
```
jshanley: is this just a way of basically returning an empty observable?

src/rx-stores/analytics.ts
```diff
@@ -1,15 +1,58 @@
...
+    } else {
+      return fetchAllObjects<DensitySpace>('/spaces');
+    }
+  }),
+  map(spaces => collectionSpacesSet(spaces)),
```
jshanley: Ah I see, so returning `of()` above short-circuits the pipe so that this action never gets emitted?

src/rx-stores/analytics.ts
```diff
@@ -1,15 +1,58 @@
...
+);
+
+merge(
+  routeTransitionStream.pipe(
+    loadSpaces(),
+  ),
+  routeTransitionStream.pipe(
+    loadSpaceHierarchy(),
+  ),
+).subscribe(action => rxDispatch(action));
```
jshanley: so, each time the route transitions to analytics, we get the spaces and hierarchy and then each corresponding collectionSet action is dispatched.

Is the intent to store the collections in the Analytics store?

src/rx-stores/index.ts
jshanley: I think this answered my earlier question. So this is where the spaces and hierarchy will live?

src/rx-stores/index.ts
```diff
@@ -87,5 +91,9 @@ export function rxLegacyReduxMiddleware() {
 
 // LEGACY: save the redux store for interop
 export function rxLegacyReduxSetStore(store) {
+  if (rxStoreSubscriptionUnsubscribe) { rxStoreSubscriptionUnsubscribe(); }
   reduxStore = store;
+  rxStoreSubscriptionUnsubscribe = reduxStore.subscribe(() => {
+    RxReduxStore.next(reduxStore.getState());
+  });
```
jshanley: does the callback for `.subscribe` here get the current state as an argument?
eg. could you avoid the `getState` call by doing:
```
reduxStore.subscribe(value => {
  RxReduxStore.next(value)
}
```

src/types/redux.ts
jshanley: I can jump in and help on the typings here once I get in an editor. In some contexts you can help the compiler by adding `as const` when you mean the literal string. For example:

``` typescript
const thingA = {
  type: 'A'
}

const thingB = {
  type: 'B' as const
}

// this is allowed because "typeof thingA" resolves to {type: string}
const newThing: typeof thingA = {
  type: 'C'
}

// this is an error because "typeof thingB" resolves to {type: 'B'}
const newThing: typeof thingB = {
  type: 'C'
}

src/rx-stores/analytics.ts
1egoman: Yes

src/rx-stores/analytics.ts
1egoman: Correct!

src/rx-stores/index.ts
1egoman: It does not :(

src/rx-stores/analytics.ts
1egoman: When we talked before, we came to the conclusion (as I understand it) that they wouldn't be stored in the analytics store. I kinda think they should be though.

src/rx-stores/index.ts
1egoman: Possibly, see my other comment on this...

PR #364, opened 2019-09-09
Live buildings

rjgrazioli:
We have a number of requests from customer success to start displaying live count tickers for buildings and floors. It's been impacting a few accounts.

I removed this space-type filters and replaced them with a "has doorways" filter. This way parent spaces like buildings won't show up unless they're actually being measured.


rjgrazioli:
What's with these package conflicts? I always get these. Is there any way I can avoid them?

src/components/live-space-list/index.tsx
```diff
@@ -56,7 +56,9 @@ export function LiveSpaceList({
   }
 
   // Remove campuses, buildings, and floors before rendering.
```
1egoman: Can you update this comment? With the change you made below it's no longer correct.

src/components/live-space-list/index.tsx
```diff
@@ -56,7 +56,9 @@ export function LiveSpaceList({
...
-  filteredSpaces = filteredSpaces.filter(i => i.spaceType === 'space');
+  filteredSpaces = filteredSpaces.filter(i => i.doorways.length !== 0);
+
+  console.log(filteredSpaces);
```
1egoman: Can you remove the `console.log`?

PR #369, opened 2019-09-11
Stage 1: RxStore modal and attempting to type thunks

1egoman:


src/components/account/index.tsx
```diff
@@ -465,7 +467,7 @@ export default connect((state: any) => {
 }, dispatch => {
   return {
     async onUpdateAlert(alert) {
-      await collectionAlertsUpdate(dispatch, alert);
+      await collectionAlertsUpdate(dispatch as DispatchType, alert);
```
1egoman: This isn't technically correct, since the `dispatch` being referred to here is redux's dispatch, and not `rxDispatch`. If there's a better way to do this, I'd love to hear about it.

src/components/admin-user-management/index.tsx
```diff
@@ -95,8 +95,8 @@ export function AdminUserManagement({
         <Modal
           visible={activeModal.visible}
           width={783}
-          onBlur={() => dispatch(hideModal() as any)}
-          onEscape={() => dispatch(hideModal() as any)}
+          onBlur={() => hideModal(dispatch)}
```
1egoman: I made a rx-actions interface for the modal and ported at least the user management stuff to use it. I did this first in this pull request when I didn't realize that the thunk stuff was related, but I think it's still a worthwhile change.

src/components/admin-user-management/index.tsx
```diff
@@ -193,14 +193,14 @@ export function AdminUserManagement({
           />
         </AppBarSection>
         <AppBarSection>
-          <Button type="primary" variant="filled" onClick={() => dispatch(
-            showModal('MODAL_ADMIN_USER_ADD', {
+          <Button type="primary" variant="filled" onClick={() => {
+            showModal(dispatch, 'MODAL_ADMIN_USER_ADD', {
```
1egoman: Why is `'MODAL_ADMIN_USER_ADD'` a string? Shouldn't this have a corresponding constant somewhere in `src/actions`?

src/rx-stores/index.ts
```diff
@@ -76,7 +77,12 @@ export function rxDispatch(action: GlobalAction) {
 
   // LEGACY: dispatch every action to redux too
   if (reduxStore) {
-    const reduxAction = { ...action, legacyReduxAction: true };
+    let reduxAction;
```
1egoman: This is the bit of logic I added last week that caused these issues. I was encountering an issue where actions that were dispatched twice to all rx stores, because the action would do this:
1. First, it would be dispatched by `rxDispatch`
2. Next, it would be processed by `rxDispatch`, and be dispatched to the action stream (causing it to be processed by all stores)
3. Then, it would be dispatched to redux
4. Then, it would be handled by the redux middleware, which re-dispatches it to all rx stores.
5. And this then dispatches it a second time to all the rx-stores.

So to work around this, I added a parameter to all redux actions that were dispatched so that they wouldn't be double-dispathced by the above problem. But since `GlobalAction` didn't contain a `function`, the types didn't cause this to error.

jshanley: @1egoman I think this is working now. 

I flipped around the logic so that when dispatching via `rxDispatch`:

1. If the action is a thunk:
    - it is NOT passed along to the Rx action stream
    - it is immediately passed through to Redux
2. If the action is NOT a thunk:
    - it is passed along to the Rx action stream
    - it is then wrapped in a special action that indicates that it has already been handled in Rx-land
    - the wrapped action is passed through to Redux

...then in the Redux middleware:
1. If the action is a thunk:
    - next(action)
2. If the action is in the special wrapper:
    - next(action.action) _basically unwrap the action and pass it along_
3. Otherwise it hasn't been handled by Rx yet:
    - pipe the action into the Rx action stream
    - next(action)

guscost:
Closing, change are in #389

src/components/account/index.tsx
guscost: A better way is to finish migrating from redux!

PR #392, opened 2019-09-19
Analytics: hover state and tooltip updates

jshanley:
Resolves [AN-209](https://densityio.atlassian.net/browse/AN-209)

Changes:
- fixed bug where date tooltip displayed midpoint time of bucket rather than start time
- limited tooltip to 5 closest values to cursor
- added label of how many spaces are hidden when there are more than 5 to be highlighted
- mouse target now covers the whole chart area
- updated story for `analytics-line-chart-tooltip` to test behavior


jshanley:
resolves [AN-209](https://densityio.atlassian.net/browse/AN-209)

PR #400, opened 2019-09-23
Analytics - tests and perf updates

jshanley:


src/components/analytics-line-chart/helpers.ts
```diff
@@ -9,23 +9,57 @@ import { Bucket, HoverStateValid } from './types';
...
+
+    // ...and also to group them by local start time
+    let group: AnalyticsDatapoint[];
+    if (datapointGroups.has(datapoint.startLocalEpochTime)) {
+      group = (datapointGroups.get(datapoint.startLocalEpochTime) || [])
+    } else {
+      group = []
+    }
+    group = group.concat([datapoint])
+    datapointGroups.set(datapoint.startLocalEpochTime, group)
```
guscost: I think this makes sense, "local epoch time" is a weird concept but it is named clearly.

src/components/analytics-line-chart/index.tsx
```diff
@@ -342,4 +345,12 @@ const AnalyticsLineChart: React.FunctionComponent<{
   )
 }
 
-export default AnalyticsLineChart;
+export default React.memo(AnalyticsLineChart, (prevProps, nextProps) => {
```
guscost: whoa neat, shouldComponentUpdate?

jshanley: yep

PR #408, opened 2019-09-23
Fix org week start logic for relative ranges

jshanley:
Resolves [AN-252](https://densityio.atlassian.net/browse/AN-252)


src/helpers/space-time-utilities/index.ts
```diff
@@ -170,26 +170,49 @@ export function realizeDateRange(
...
-    timestamp = timestamp.add(DAYS_OF_WEEK.indexOf(organizationalWeekStartDay), 'days');
+    let targetDay = DAYS_OF_WEEK.indexOf(organizationalWeekStartDay);
+    // FIXME: this shouldn't be possible, but default to Sunday if input is bad/undefined
+    if (targetDay === -1) { targetDay = 0; }
+    
+    // Week should start on the target day, and end at the end of 1 before the target day
+    //  eg. [Wednesday @ 00:00, Tuesday @ 23:59] if target day is Wednesday
+    let targetDayWeekStart = targetDay;
+    // silly JavaScript modulo of negative numbers, but this is basically just mod(7)
+    let targetDayWeekEnd = (((targetDay - 1) % 7) + 7) % 7;
```
1egoman: I don't fully understand what this does, but from the description above could this be `Math.abs(targetDay - 1) % 7`?

jshanley: Unfortunately it is not. 😭 
[Here's a post](https://web.archive.org/web/20090717035140if_/javascript.about.com/od/problemsolving/a/modulobug.htm) with some more context on this if you're curious

guscost: 😱

PR #427, opened 2019-09-26
Make space mapping button not clickable for non-active integrations

1egoman:
That being said, right now there isn't a way to make a cell in a `ListView` optionally clickable, so I made it so if one clicks on it, nothing happens (but `cursor: pointer` is still set on the `ListView` cell). I held off on making changes to the list view in regards to this so that we could talk about it as a group.


guscost:
Merging this, but will add a note to review the list-view component and this behavior with everyone.

PR #431, opened 2019-09-27
unstable: replace user reducer enhancer with side chain

jshanley:


src/rx-stores/user/index.ts
```diff
@@ -36,4 +45,8 @@ export function user(state=initialState, action) {
   }
 }
 
-export default analyticsUserReducerEnhancer(user);
+const UserStore = createRxStore('UserStore', initialState, userReducer);
+
+actions.pipe()
```
1egoman: What is this empty pipe for?

jshanley: oops, nice catch

PR #497, opened 2019-10-16
Firmware version and filtering (!) in DPU status page

bredfield:


src/components/admin-device-status/index.tsx
```diff
@@ -46,63 +50,82 @@ export function AdminDeviceStatus({
...
-      </ListView>
-    </div>
-  </AppScrollView>;
+
+  return <Fragment>
+    <AppBar>
+      <InputBox
+        type="text"
+        width={250}
+        className={'hey'}
```
jshanley: 👋

PR #528, opened 2019-10-28
DASH-191, DASH-323, DASH-326, DASH-327

guscost:
Several smaller issues from the Location Management 1.1 epic.

As part of DASH-323, this build consolidates common snippets that are re-used across the four space types. These snippets are in a new `admin-locations-snippets` module in the components folder.


src/types/index.ts
```diff
@@ -28,6 +28,11 @@ export enum DensitySpaceFunction {
   // Note, "OTHER" is not in here since typescript does not allow "null" in enums
 }
 
+export enum DensitySizeAreaUnitTypes {
+	SQUARE_FEET = 'square_feet',
```
1egoman: There are tabs here, fyi

src/components/admin-locations-snippets/index.tsx
```diff
@@ -0,0 +1,234 @@
+import React, { Fragment } from 'react';
+import classnames from 'classnames';
+import comma from 'comma-number';
```
jshanley: You could also just use `Number.prototype.toLocaleString` and use `en-us`:

eg. 

``` javascript
var n = 1234
n.toLocaleString('en-us') // '1,234'
```

src/components/admin-locations-snippets/README.md
```diff
@@ -0,0 +1,12 @@
...
+This module has lots of snippets used to render location admin pages.
+
+## AdminLocationsLeftPaneDataRow
+
+A component used to render a dataset in the left hand pane of space management list views. This
+component is a wrapper for a metric group which is displayed in one row.
+
+## AdminLocationsLeftPaneDataRowItem
+
+This component is the "child" component for each metric, and should be nested inside an AdminLocationsLeftPaneDataRow component.
```
jshanley: super helpful, thank you 🙌

src/components/admin-locations-snippets/index.tsx
guscost: unfortunately:

```
(1234.00).toLocaleString('en-us')
"1,234"
(1234.00).toFixed(2).toLocaleString('en-us')
"1234.00"
```

jshanley: (someNumber).toFixed(2) is a String, not a Number

guscost: the point is this method is unusable for converting a number of dollars and cents to a comma-formatted string

PR #537, opened 2019-10-31
New OKTA Workflow with a static "LinkedIn" [DASH-374]

rjgrazioli:
- I decided to just keep it simple and add the okta option
- I haven't tested on mobile yet but it should just work
- I wanted to get this PR started to ensure you guys were cool with the workflow
- Might add some design elements before we ship, but I'd ship as is
- Ask Canute to set you up with an account if you want to test okta


beweinreich:
This might be a stupid question... but how do I test this?

PR #610, opened 2019-11-29
[imp-90] Low power mode support

bredfield:
Adds visibility into DPUs that are in low power mode, and prompts for remedy when 1 or more fall into that category. 

image

@rjgrazioli I took at stab at the styling for the warning pane, feel free to let me know how 
I should improve it.

Coupled with DensityCo/dashboard#610 (but can be deployed in any order)

### DIFF
--- a/src/components/admin-device-status/index.tsx
+++ b/src/components/admin-device-status/index.tsx
@@ -104,13 +104,13 @@ export function AdminDeviceStatus({
     <div className={styles.lowPowerWarningSection}>
       <div className={styles.lowPowerWarningAlert}>
         <Icons.Danger color='#f4ab4e'/>
-        <h3 className={styles.lowPowerWarningTitle}>
+        <h4 className={styles.lowPowerWarningTitle}>
           {numLowPowerSensors} DPUs are not counting due to insufficient power. Please&nbsp;
           <span
             className={styles.lowPowerWarningLink}
             onClick={openIntercom}>reach out to customer support.
           </span>
-        </h3>
+        </h4>
       </div>
     </div>
   ) : null;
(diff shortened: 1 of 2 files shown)


src/components/admin-device-status/index.tsx
```diff
@@ -51,15 +75,38 @@ export function AdminDeviceStatus({
   spaces
 }) {
   const [ sensorsFilterText, setSensorsFilterText ] = useState('');
-
   let sortedSensors = sensorsFilter(sensors.data, sensorsFilterText).sort(function(a, b){
-   if (a.status < b.status)
+    if (a.status === 'low_power') {
+      return -1;
+    }
```
guscost: Mixing curly-braces and no-curly-braces style here... just do curly braces everywhere?

rjgrazioli:
Yowza, that's a lot of copy @bredfield 

Is there a way we could break that up to be more digestible? Perhaps starting with an alert and include an action that opens a modal with more information?

power bannner

Screen Shot 2019-12-02 at 12 56 00

I have better guidelines on these UI elements set to release next week too. FYI

src/components/admin-device-status/index.tsx
jshanley: definitely curly braces when using a line break IMO.. eg.

this is fine:
``` javascript
if (a.status < b.status) return -1
```

but, not a fan of this:
``` javascript
if (a.status < b.status)
  return -1;
```

bredfield:
@rjgrazioli Yeah, we can do that. Will see if I can get that in this afternoon with the current UI tooling

bredfield:
@rjgrazioli I think for right now I'm going to remove the detail text and will add a link to a helpdesk article once CX gets that up.

bredfield:
@rjgrazioli 
image

bredfield:
Screen Shot 2019-12-03 at 8 29 07 AM

PR #631, opened 2019-12-11
Hide spaces without capacity when viewing utilization metrics

jshanley:
Resolves [DASH-606](https://densityio.atlassian.net/browse/DASH-606)


src/components/analytics-chart/index.tsx
```diff
@@ -930,6 +930,21 @@ const AnalyticsChart: React.FC<{
...
+      const additionalHiddenSpaceIds = report.queryResult.data.selectedSpaceIds.filter(spaceId => {
+        const space = spacesById.get(spaceId);
+        if (!space) return false;
+        if (space.targetCapacity == null) return true;
+        return false;
+      });
+      return report.hiddenSpaceIds.concat(additionalHiddenSpaceIds);
+    }
+    return report.hiddenSpaceIds;
+  })()
```
guscost: Sounds like a good idea, but I don't get this functional-but-imperative-inline-IIFE-with-no-arguments thing.

jshanley: Oh yeah, let's talk about it

PR #632, opened 2019-12-11
include doorway environment in doorway GET

guscost:
Should fix DASH-585 with no backend changes needed.


src/rx-actions/route-transition/admin-brivo-mappings.ts
```diff
@@ -39,7 +39,7 @@ async function fetchAllBrivoSites(dispatch) {
   try {
     // BRIAN: Not sure how guys handle loading states, but I'm using this to determine if it's loading...
     dispatch({type: 'BRIVO_SITES_LOADING', sitesLoading: true})
-    doorways = await fetchAllObjects<DensityDoorway>('/doorways', { cache: false });
+    doorways = await fetchAllObjects<DensityDoorway>('/doorways?environment=true', { cache: false });
     doorwayMappings = await fetchAllObjects<DensityDoorway>('/integrations/doorway_mappings/', { cache: false });
```
jshanley: doorway mappings are also type `DensityDoorway`?

src/rx-actions/route-transition/admin-locations-edit.ts
```diff
@@ -26,7 +26,7 @@ export async function loadData(dispatch) {
     [hierarchy, spaces, doorways, labels, tags, assignedTeams] = await Promise.all([
       await fetchAllObjects<DensitySpaceHierarchyItem>('/spaces/hierarchy', { cache: false }),
       await fetchAllObjects<DensitySpace>('/spaces', { cache: false }),
-      await fetchAllObjects<DensityDoorway>('/doorways', { cache: false }),
+      await fetchAllObjects<DensityDoorway>('/doorways?environment=true', { cache: false }),
```
jshanley: what does this query param do?

jshanley: oh that's the whole point ;)

jshanley: Looking at the implementation of `fetchAllObjects` it looks like you can pass the additional query parameter like this:

``` typescript
fetchAllObjects<DensityDoorway>('/doorways', {
  cache: false,
  params: {
    environment: 'true',
  }
})
```

src/rx-actions/route-transition/admin-brivo-mappings.ts
guscost: environment mappings? yes those are part of the typedef

src/rx-actions/route-transition/admin-locations-edit.ts
guscost: nice, thanks, will change

src/rx-actions/route-transition/admin-brivo-mappings.ts
guscost: oh, no idea about doorway mappings... it probably turns into an `any` type implicitly anyway

PR #633, opened 2019-12-12
hotfix for available capacity feature flag

jshanley:


src/components/analytics/index.tsx
```diff
@@ -148,7 +148,9 @@ export default function Analytics() {
     throw new Error('This should not be possible')
   }
 
-  const opportunityMetricEnabled = Boolean(getInObject(user.organization, 'settings.opportunityMetricEnabled', false));
+  // incredible fumble, these flags coming from core are all strings eg. 'true' or 'false', wow
+  const opportunityMetricEnabledString = getInObject(user.organization, 'settings.opportunityMetricEnabled', 'false');
```
guscost: is the implementation of `getInObject` basically `object[key] === undefined ? default : object[key]` ?

guscost: oh it's nested n levels, cool

jshanley: yeah it's nice

PR #646, opened 2020-01-02
dependencies on shared (TypeScript) helpers

guscost:
This build adds dependencies on `@density/lib-*` NPM packages that contain TypeScript implementations of various helpers.

NOTE: Importing un-transpiled packages with ES6 exports is not conventional for NPM packages, so we have to override some settings that Webpack and Jest use when running, building, and testing the project. I'd rather do this for now than come up with a build system and pretty much lose all TS sourcemaps when developing, but we should plan to change this later.

NOTE: The @density/lib-api-types package is just TypeScript types. If we transpile this for NPM, it would not even exist anymore, right? Let's talk about what to do here. I agree with @jshanley that a shared repo for types would be a great resource, even if we have to manage them by hand. I think it's worth it even if this package has to permanently break with NPM convention.


guscost:
verbal approval obtained.

PR #651, opened 2020-01-03
import raw src files from density ui

guscost:


src/components/account-registration/index.tsx
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 
 import { DensityUser } from '../../types';
 
-import { Button, InputBox } from '@density/ui';
+import { Button, InputBox } from '@density/ui/src';
```
jshanley: @guscost how does this work, does the `src` folder have an index file? What if you wanted just a specific module, but not the whole library?

config-add-paths.js
```diff
@@ -13,6 +13,7 @@ module.exports = function addPaths(rule) {
       path.resolve('./node_modules/@density/lib-helpers'),
       path.resolve('./node_modules/@density/lib-space-helpers'),
       path.resolve('./node_modules/@density/lib-time-helpers'),
+      path.resolve('./node_modules/@density/ui/src'),
```
jshanley: does this still work the same way if you remove `/src` from the end of that resolve path?

src/components/account-registration/index.tsx
guscost: The src file has an index, yes, which imports all the other modules.

I believe you can also do e.g. `import AppBar from '@density/ui/src/app-bar'`, but you have to do separate lines for each file you're importing from in that case.

config-add-paths.js
guscost: It breaks, the react import is mysteriously undefined.

guscost:
Interestingly, it gets a bit bigger:

```
File sizes after gzip:

  915.39 KB (+74.88 KB)  build/static/js/2.c05bbd6f.chunk.js
  150.26 KB (-518 B)     build/static/js/main.73c95f43.chunk.js
  26.43 KB               build/static/css/main.c78b336e.chunk.css
  15.43 KB (+11.12 KB)   build/static/css/2.6f47ce3a.chunk.css
  782 B                  build/static/js/runtime-main.cbf131a3.js
```

guscost:
However, this is after replacing imports from the bundled package with imports from the package index file as a .ts, if we import directly from individual source files it might shake the tree a bit better.

PR #679, opened 2020-01-13
Refactor API types to keep snake-case

guscost:
Renamed properties from API types to avoid the need for camelCase conversion. Updated @density/reports to 12.1.0 to match these changes. Some other properties have been converted to snake_case unnecessarily, but this is ok for the short term.


guscost:
Did a fairly extensive manual test of all critical dashboard features.

PR #735, opened 2020-02-01
PRD-103 / Adds support link to navbar

rjgrazioli:
Add a support link to the navbar that opens in a new tabs so that customers know our support center actually exists.


src/components/app-navbar/index.tsx
```diff
@@ -206,6 +200,16 @@ export default function AppNavbar({
             text=""
           />
 
+          <AppNavbarItem
+            selected={false}
+            showOnMobile={true}
+            path="https://www.density.io/support/"
+            newTab={true}
```
guscost: Sort of a matter of taste, but I'd call this `targetBlank`

PR #794, opened 2020-03-06
Refactor "dashboard reports" date range behavior

guscost:
Migrates the analytics "date range" logic into `@density/lib-time-helpers`, and uses a new version of the reports repo that shares this logic.

Also modifies a convention so that reports are rendered with respect to "yesterday", not the start of the latest week. Solves PRD-97.


src/rx-actions/dashboards/report-modal.ts
```diff
@@ -64,7 +64,10 @@ export async function rerenderReportInReportModal(
 ) {
   clearPreviewReportData(dispatch);
 
-  const currentDate = moment(dashboardDate || undefined).format('YYYY-MM-DD');
+  // TODO: until we have designs for "partial days", dashboards default to yesterday
+  const currentDate = dashboardDate ?
+    moment(dashboardDate).format('YYYY-MM-DD') :
+    moment().subtract(1, 'day').format('YYYY-MM-DD');
```
rjgrazioli: You're a real hero for this gus, haha

rjgrazioli:
All of the UI components look ok to me, only rodeo has no dummy data anymore so I can't really test any data. But my guess is it's fine to ship. Good step forward with the time helpers library.

Mike is well on his way to revamping time segments, then we'll be off to the races with reports. Hang in there

PR #812, opened 2020-03-23
Two overnight reports for "spaces" page

guscost:
This updates the "hourly breakdown" and "periodic metrics" reports so they are compatible with time segments that go "overnight".


guscost:
Been doing manual QA of this one, since nobody understands it.

PR #817, opened 2020-03-30
Tailgating webhooks

1egoman:


src/components/admin-developer/index.tsx
```diff
@@ -67,25 +65,61 @@ export function TokenKeyHider ({value, onCopyToken}) {
...
-  onCloseModal,
-  onCopyToken,
-}) {
-  return <Fragment>
+  const onOpenModal = (name, data) => {
+    showModal(dispatch, name, data);
+  }
+  const onCloseModal = async () => {
+    await hideModal(dispatch);
+  }
```
guscost: I'd probably inline these, to limit indirection (there's another use of `showModal` directly anyway)

src/components/admin-developer/index.tsx
```diff
@@ -67,25 +65,61 @@ export function TokenKeyHider ({value, onCopyToken}) {
...
+
+  const onCreateToken = async (token) => {
+    const ok = await collectionTokensCreate(dispatch, token);
+    if (ok) { hideModal(dispatch); }
+  }
+  const onUpdateToken = (token) => {
+    updateModal(dispatch, { token });
+  }
```
guscost: Ehhh this one is a single-liner too, I'm on the fence about inlining but still would lean slightly towards it. No big deal.

src/rx-actions/collection/webhooks/create.ts
```diff
@@ -1,23 +1,28 @@
...
       name: item.name,
       description: item.description,
       endpoint: item.endpoint,
+      headers: item.headers,
     });
-    dispatch(collectionWebhooksPush(response.data));
-    return response.data;
   } catch (err) {
-    dispatch(collectionWebhooksError(err));
+    showToast(dispatch, { type: 'error', text: 'Error creating webhook.' });
```
guscost: Love it here, less indirection!

src/components/admin-webhook-modal/index.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+type Headers = {[name: string]: string};
+
+type WebhookHeaderControlProps = {
+  headers: Headers,
+  onChange: (headers: Headers) => void,
+  disabled: boolean,
+};
+
+const NEW_ROW = 'NEW_ROW';
```
guscost: Academically, I'd like to debate whether `null` implicitly meaning this would be better or worse, but definitely not going to object. We should come up with a "new id" convention for the whole dashboard though.

src/components/admin-webhook-modal/index.tsx
```diff
@@ -0,0 +1,311 @@
...
+                <p>
+                  A webhook will be sent as a POST request to the URL specified in JSON whenever
+                  a tailgating event at any of your doorways occurs.
+                </p>
+                <p>Here's an example tailgating event webhook payload:</p>
+                <pre className={styles.webhookCreateExamplePayload}>{
+                  JSON.stringify({
+                    "type": "tailgating_event",
+                    "request_id": "5245250ce4a447e18f513b1863c80ea0",
+                    "version": "v3",
```
guscost: Oooh, v3 has been broken open!

src/components/admin-webhook-modal/index.tsx
```diff
@@ -0,0 +1,311 @@
...
+                    "payload": {
+                      "trajectory": -1,
+                      "doorway_id": "drw_643916192641188838",
```
guscost: Why not the same doorway as the above example? Just curious, what are these spaces/doorways?

1egoman: I'm not sure actually, that was just an id that was in a document brian had.   I'll update it to match the doorway id already in the other example.

PR #827, opened 2020-03-31
Added in roadmap tab

beweinreich:
Added in a roadmap tab that will render an iframe.

### DIFF
--- a/src/components/app/index.tsx
+++ b/src/components/app/index.tsx
@@ -29,6 +29,7 @@ import impersonateSet from '../../rx-actions/impersonate';
 import { defaultState as impersonateDefaultState } from '../../rx-stores/impersonate';
 
 import Dashboard from '../dashboard';
+import Roadmap from '../roadmap';
 import AppNavbar from '../app-navbar';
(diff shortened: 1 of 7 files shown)


beweinreich:
Waiting on URL from Andrew

src/components/app/index.tsx
```diff
@@ -179,6 +180,8 @@ function ActivePage({activePage, user, settings}) {
     return <DashboardsEdit />;
   case "ANALYTICS":
     return <Analytics />;
+  case "ROADMAP":
+    return stringToBoolean(settings.roadmap_enabled) ? null : <Roadmap />;
```
guscost: Is this backwards?

rjgrazioli:
Does it need to be a top level name item? Can we nest it under the Admin section?

guscost:
Fixed what looks like a bit of backwards logic, and added some sandboxing rules to tighten down what the iframe is allowed to do.

beweinreich:
thanks @guscost 

also dont merge yet (!) still need the url from andrew

PR #839, opened 2020-04-16
PRD-723: Resolve CSV download vulnerabilities

jshanley:
Added a sanitize function that takes the text of a CSV document and enforces the following rules:
  - only alphanumeric characters and spaces are allowed in cells
  - except: minus signs are allowed, but only before numeric digits

Applied this function in all places that need it (please help check me on this)


src/helpers/csv/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+
+const safeDocument = `one,two,three
+1,2,3
+4,5,6
+7,8,9
+-1,-2,-3`;
+
+const dangerDocument = `one,=two,three
+@1,!2,+3
```
1egoman, line 11: It might be good to use the example they used in the pentest, or at least have it in here somewhere.

jshanley: Good idea

PR #840, opened 2020-04-16
code: call logout in API when logging out

guscost:


src/rx-actions/route-transition/logout.ts
```diff
@@ -5,11 +5,13 @@ import collectionTokensSet from '../collection/tokens/set';
 import collectionWebhooksSet from '../collection/webhooks/set';
 import collectionServicesSet from '../collection/services/set';
 import { UserActionTypes } from '../../types/users';
+import accounts from '../../client/accounts';
 
 export const ROUTE_TRANSITION_LOGOUT = 'ROUTE_TRANSITION_LOGOUT';
 
 export default function routeTransitionLogout(dispatch) {
   dispatch({ type: ROUTE_TRANSITION_LOGOUT });
+  accounts().delete('/logout');
```
jshanley, line 14: Sending a DELETE request to `/logout`? Also, what if this fails?

guscost: Either we say "error logging out" and the user is stuck logged-in forever, or we discard the token on the frontend anyway. Perhaps the former is more robust.

guscost: DELETE is the correct verb, I don't want to alter the API.

jshanley: 10-4, just making sure that was what you meant to do

jshanley: I think we need to handle a failed request somehow. In the context of the pen test, you could do all the same steps as mentioned, then click "Turn WiFi Off", and click logout, etc. etc. and the test would fail the same as before. (Or drive under a bridge, or the API is down for a second, etc..)

jshanley: I think the bottom line is.. if the user isn't logged out, the UI needs to reflect that. To me, having an error that says "failed to log out" might be irritating to the user, but it would at least reflect the correct state of the world and allow the user to decide how they want to handle that situation.

guscost: OK sounds good, I'll add an error toast and interrupt the logout in this case.

PR #841, opened 2020-04-17
Allow timestamps to pass sanitization

guscost:


jshanley:
@guscost I'm making this more explicit.. now it strips whitespace first, then looks for leading control characters and gobbles them up. If you're alright with this, let's ship it. Any concerns, note them here and assign back to me.

PR #850, opened 2020-05-04
Onprem integration

1egoman:
- Adds a new constant that allows a number of ui elements to be shown / hidden when the dashboard is deployed in "on prem mode"
- Adds the "error collector", a system to log errors to a locally running server when the dashboard is built in on prem mode.


package.json
```diff
@@ -19,6 +19,7 @@
     "@density/ui": "^21.5.0",
     "@mapbox/mapbox-gl-geocoder": "^4.5.0",
     "@material-ui/core": "^4.7.1",
+    "@sentry/browser": "^5.15.5",
```
guscost, line 22: Is there a reason we're using sentry's tools here? What is the on-prem server that receives these? Are we using sentry.io with on-prem installations?

src/error-collector.ts
```diff
@@ -0,0 +1,192 @@
...
+
+// Include these fields in errors when they are sent to the server
+const ERROR_FIELDS = {
+  locationHref: () => window.location.href,
+  browserTimestamp: () => moment().format(),
+  userAgent: () => window.navigator.userAgent,
+  storeStateSnapshots: () => {
+    const snapshots = {};
+    for (let displayName in storeDirectory) {
+      snapshots[displayName] = storeDirectory[displayName].imperativelyGetValue();
```
guscost, line 22: This will send secrets/tokens to the log server.

src/fields.ts
```diff
@@ -43,3 +43,6 @@ export default [
     default: process.env.REACT_APP_ENVIRONMENT || 'production',
   },
 ];
+
+export const ON_PREM = false;
```
guscost, line 47: How will this ever be true??? This PR is likely incomplete.

src/on-prem/styles.scss
```diff
@@ -0,0 +1,156 @@
+@import "@density/ui/variables/colors.scss";
```
guscost, line 1: Not over the moon about a random folder called "on-prem", but this works.

src/rx-stores/user/effects.ts
```diff
@@ -21,10 +23,12 @@ export default function registerSideEffects(
...
 
-    // skip tracking if user is impersonating
-    if (isUserImpersonating()) return;
+    // Skip tracking if user is impersonating
+    if (isUserImpersonating()) { return; }
+    // Also skip tracking if this is an on-prem build
+    if (ON_PREM) { return; }
 
-    if (process.env.REACT_APP_HOTJAR_SITE_ID) {
+    if (process.env.REACT_APP_HOTJAR_SITE_ID && !ON_PREM) {
```
guscost, line 31: Redundant with the above short-circuit.

package.json
guscost: I'm not at all in favor of adding new dependencies that we don't need, much less implicitly adding downstream packages (as needed by error-collector.ts) like this. Just spent about an hour today sorting out why another implicit import broke when I tried to update sinon.

src/error-collector.ts
1egoman: Correct, it sounds like you're suggesting it should not? I can probably skip the stores that contain this info.

src/fields.ts
1egoman: Mentioned this yesterday but repeating here for others: currently, this file is completely replaced by the on prem build process. Not very satisfied with that decision but that's why there's no way to "turn it on".

src/error-collector.ts
1egoman: Done

package.json
1egoman: Good point, I hadn't realized I had forgotten to make those dependencies. Done.

1egoman:
No longer relevant

PR #900, opened 2020-05-28
Migrate navbar to point to prototype

1egoman:


src/components/app-navbar/index.tsx
```diff
@@ -175,7 +175,7 @@ export default function AppNavbar({
           {stringToBoolean(settings.queue_enabled) ? <AppNavbarItem
             selected={['QUEUE_SPACE_LIST', 'QUEUE_SPACE_DETAIL'].includes(page)}
             showOnMobile={true}
-            path="#/display/spaces"
+            path="https://safe.density.io/#/displays"
```
1egoman, line 178: NOTE: we need to figure out how to swap in staging here eventually

guscost: 👍

PR #904, opened 2020-06-01
Remove queue from dashboard, redirect to display prototype instead

1egoman:


src/helpers/generate-display-prototype-url/index.ts
```diff
@@ -0,0 +1,12 @@
+// Given a path, generate a link to the display prototype
+export default function generateDisplayPrototypeUrl(path, href=window.location.href) {
```
1egoman, line 2: Only slightly complicated part here is this - I wrote a helper to figure out which version of the prototype to redirect to based on the current url. I _think_ this is a good idea - maybe if this lives for a while and we're able to do some work on the environment switcher, then we could make this an option in there that defaults to something like this.

bredfield: I think this is sane, and thanks for the tests

PR #906, opened 2020-06-03
Add npm script to check for implicit anys

jshanley:
Added a script to run the typescript compiler with the `noImplicitAny` flag turned on. 

To use it, run:
``` sh
npm run anycheck
```

This will give us the locations and a final count of implicit anys in the codebase. This is just intended as a metric for us to track over time. We could automate the reporting of it if desired, but for now it's just something we can run periodically to give ourselves a sense of how we're tracking towards the eventual goal of eliminating all untyped code.

Note: For a meaningful count, this should be run once `npm run typecheck` gives a zero exit code, otherwise it will catch other non-implicit-any errors as well.

Note: The script is piping the tsc output to `true` to suppress npm errors, since this check is expected to give a non-zero exit code until we eliminate all implicit anys.

If/when we are ready to make the jump to `noImplicitAny` in the actual tsconfig, this script can be removed.


jshanley:
@guscost I want to be able to do single files also, but it's surprisingly difficult to get that to work. See this thread for some color: https://stackoverflow.com/questions/44676944/how-to-compile-a-specific-file-with-tsc-using-the-paths-compiler-option

Merging this for now. I'll keep exploring options for single file checking.

PR #913, opened 2020-06-08
New space management w/ composite spaces

guscost:
Updates the "space management" admin page to a new layout, and adds support for the "composite spaces" counting mode.

Please test:

- Updating each management section for existing spaces
- Updating each management section for new spaces
- Turning on the `composite_spaces_enabled` flag and testing the new "Data Settings" section
- Testing the new "Component Spaces" list on the Spaces page for a composite space
- Testing the stock Confirm/Prompt interactions that use the new Dialogger (cc @1egoman)


src/components/admin-locations-detail-modules/metadata.tsx
```diff
@@ -72,8 +72,8 @@ export default function AdminLocationsDetailModulesMetadata({space_type, formSta
           ]}
           disabled
           width={158}
-          value={formState.currencyUnit}
-          onChange={choice => onChangeField('currencyUnit', choice.id)}
+          value={formState.currency_unit}
+          onChange={choice => onChangeField('currency_unit', choice.id)}
```
guscost, line 76: I think this one was just broken before.

src/index.tsx
```diff
@@ -474,11 +483,19 @@ spacesPageEventSource.on('space', countChangeEvent => {
...
+        state.liveEvents.metrics.entrances,
+        state.liveEvents.metrics.exits,
+      ));
+    } else {
+      rxDispatch(spacesPageActions.setLiveStats(
+        Math.max(state.liveEvents.metrics.occupancy + countChangeEvent.direction, 0),
+        state.liveEvents.metrics.entrances + (countChangeEvent.direction > 0 ? countChangeEvent.direction : 0),
+        state.liveEvents.metrics.exits + (countChangeEvent.direction < 0 ? -1 * countChangeEvent.direction : 0),
+      ));
+    }
```
guscost, line 498: This stuff is sort of a half-measure to attempt to support count resets. Much more extensive refactoring, which removes all of this, is in another branch.

rjgrazioli:
- [ ] The nav items should be 40px tall
- [ ] The inactive nav items are a medium font weight
- [ ] Truncate the long breadcrumbs—or go back to the drawing board because this design will break a lot

This is all I have assuming we'll make more of the updates later

guscost:
Yes, the rest of the changes should come afterwards, this is just so that we don't block the composite spaces feature.

I'm not sure about the last note, the breadcrumb should line-break at each item.

rjgrazioli:
> I'm not sure about the last note, the breadcrumb should line-break at each item.

If that's the case, then I don't like this design. That head is going to get awkwardly tall. Can you address with @kyleohara

src/components/admin-locations-detail-modules/component-spaces.tsx
```diff
@@ -0,0 +1,51 @@
...
+              value={countingMode === SpaceCountingMode.SUM_EXPLICIT ? selectedSpaces : []}
+              formattedHierarchy={formattedHierarchy}
+              onChange={onChange}
+            />
+          </SpacePickerContext.Provider>
+        </div>
+        <div className={styles.componentSpacesSpaceListContainer}>
+          <AdminLocationsComponentSpaceList
+            spaces={spaces}
+            spaceIds={selectedSpaces.map(x => x.space.id)}
```
jshanley, line 44: `selectedSpaceIds` might be more descriptive as the prop name here

src/components/admin-locations-detail-modules/count-calculation-diagrams.tsx
```diff
@@ -0,0 +1,1792 @@
+import * as React from "react";
+
+export function DiagramDoorwayMode() {
```
jshanley, line 3: What's the advantage to using JSX for these diagrams instead of just having an .svg loaded in an`` tag? This is a lot of "javascript code" to have an image on the page.

src/components/admin-locations-detail-modules/danger-zone.tsx
```diff
@@ -20,21 +20,23 @@ function AdminLocationsDetailModulesDangerZoneUnconnected({
   onShowConfirm: (space: CoreSpace) => void
 }) {
   return (
-    <AdminLocationsDetailModule error title="Danger Zone">
-      <div className={styles.wrapper}>
-        <div className={styles.left}>
-          <h4>Delete this space</h4>
-          <span>Once deleted, it will be gone forever. Please be certain.</span>
+    <div className={styles.padTop}>
```
jshanley, line 23: Similar to what we discussed in sync review, this is another place where inline styles seem like a better choice.. in this case the name of the style indicates that the desired styling decision is known in this context anyway...

src/components/admin-locations-detail-modules/doorways.tsx
```diff
@@ -584,16 +587,26 @@ function AdminLocationsDetailModulesDoorways({
...
+}: {
+  formState: AdminLocationsFormState,
+  activeModal: Any<FixInRefactor>,
+  selectedSpaceIdOrNull: CoreSpace['id'] | null,
+  onHideModal: () => Promise<void>,
+  onShowModal: (modalName: string, data: Object) => void,
+  onUpdateModalData: (data: Any<FixInRefactor>) => void,
+  onCreateDoorway: (data: Any<FixInRefactor>) => Promise<void>,
+  onUpdateDoorway: (data: Any<FixInRefactor>) => Promise<void>,
+  onDeleteDoorway: (doorway_id: string) => Promise<void>,
```
jshanley, line 607: oh weird.. because these functions are `async` they technically return a Promise.. that's annoying because if they were regular functions that did `doSomething().then(() => otherThing())` instead of `await doSomething(); otherThing();` the return type would be `void` which is what I usually expect for "handler" functions... anyway this is fine but I just learned something.

src/components/admin-locations-detail-modules/doorways.tsx
```diff
@@ -792,30 +804,38 @@ const ConnectedAdminLocationsDetailModulesDoorways: React.FC<Any<FixInRefactor>>
...
     await spaceManagementUpdateDoorway(dispatch, data);
-  }
-  const onDeleteDoorway = async (doorway_id) => {
+    onChange();
+  };
+  async function onDeleteDoorway(doorway_id: string) {
     await spaceManagementDeleteDoorway(dispatch, doorway_id);
-  }
+    onChange();
+  };
```
jshanley, line 827: Is this just for consistency with the two function definitions below, or is this a style convention you'd like to enforce more broadly?

src/components/admin-locations-detail-modules/index.tsx
```diff
@@ -24,27 +24,40 @@ export const AdminLocationsDetailModulesDoorways = AdminLocationsDetailModulesDo
...
 
-export default function AdminLocationsDetailModule({
+export function AdminLocationsSectionTitle({
   title,
+}: {
+  title: string
+}) {
+  return <div className={styles.adminLocationsSectionTitle}>
+    {title}
+  </div>
```
jshanley, line 34: parens for alignment would be good here

src/components/admin-locations-detail-modules/styles.module.scss
```diff
@@ -4,14 +4,25 @@
 
 .module {
   background-color: #fff;
-
-  width: 818px;
-  border: 1px solid $gray300;
-  border-radius: $borderRadiusBase;
+  width: 800px;
+  @media screen and (min-width: 1440px) {
```
jshanley, line 8: how did you arrive at 1440px? Maybe this should be a variable somewhere if it's an important breakpoint for something

src/components/admin-locations-detail-modules/styles.module.scss
```diff
@@ -4,14 +4,25 @@
...
+    width: 960px;
+  }
 }
 .moduleError { border-color: $red; }
+.moduleIncludeBorder { 
+  border: 1px solid $gray200;
+  border-radius: $borderRadiusBase;
+}
 .moduleHideOverflow { overflow: hidden; }
 .moduleBody.padding {
```
jshanley, line 18: another case where the name of the class describes the desired style

src/components/admin-locations-edit/index.tsx
```diff
@@ -35,89 +36,186 @@ import useRxStore from '../../helpers/use-rx-store';
...
 
 
+type AdminLocationsFormSection = 
+  'General' |
+  'Operating Hours' |
+  'Doorway Mappings' |
+  'Address' |
+  'Tags' |
+  'Team Assignments' |
+  'Data Settings';
```
jshanley, line 55: prefer listing with the `|` prefixed to each line

src/components/admin-locations-edit/index.tsx
```diff
@@ -35,89 +36,186 @@ import useRxStore from '../../helpers/use-rx-store';
...
+  return formState &&
+
+    // Form state has a non-empty name
+    formState.name &&
+
+    // Operating hours module is valid
+    formState.time_zone &&
+    formState.daily_reset &&
+    formState.operatingHours &&
+    formState.operatingHours.filter(i => i.label === null).length === 0;
```
jshanley, line 88: This could return any kind of falsey value (for example if `formState` is `null` this function would return `null`). Consider wrapping the expression in `Boolean(...)`

src/components/admin-locations-edit/index.tsx
```diff
@@ -35,89 +36,186 @@ import useRxStore from '../../helpers/use-rx-store';
...
+}
+
+
+function RecursiveBreadcrumb({ spaces, id }: { spaces: CoreSpace[], id: string }) {
+  const space = spaces.find(x => x.id === id);
+  const parent = space ? spaces.find(x => x.id === space.parent_id) : null;
+  return <Fragment>
+    {parent ? <span><RecursiveBreadcrumb spaces={spaces} id={parent.id} />&nbsp;{'>'}&nbsp;</span> : null}
+    {space ? <span>{space.name}</span> : null}
+  </Fragment>
```
jshanley, line 98: parentheses must be on backorder 😛

src/components/admin-locations-edit/index.tsx
```diff
@@ -129,210 +227,282 @@ export function SpaceTypeForm({
...
-      MODULES.doorways,
-      MODULES.operatingHours,
-      MODULES.tags,
-      MODULES.teams,
-      MODULES.dangerZone,
-    ],
+  // TODO: maybe extract to helper lib
+  type ValueOf<T> = T[keyof T];
+
+  const MODULES_BY_SPACE_TYPE: {
```
jshanley, line 233: This is becoming a pretty detailed little minilanguage for describing UI modules. Let's try to define this a little more formally with types etc. when we get a chance.

src/components/spaces-snippets/index.tsx
```diff
@@ -358,6 +360,11 @@ export function SpaceRightSidebar({
   const spaceTimeZone = selectedSpace.time_zone;
   const dateString = localDate.format('YYYY-MM-DD');
 
+  // Is this space a composite space?
+  // TODO: Update API types and remove the Any<FixInRefactor>s
+  const isCompositeSpace = (selectedSpace as Any<FixInRefactor>).counting_mode === 'sum_explicit';
```
jshanley, line 365: isn't there an `enum` for the counting modes?

src/components/admin-locations-detail-modules/count-calculation-diagrams.tsx
guscost: Um, same reason Icons are JSX? Not entirely sure. Didn't try it, maybe our webpack setup works fine with importing svg?

src/components/admin-locations-detail-modules/danger-zone.tsx
guscost: Yeah, this is getting toward "functional CSS"/tailwind style anyway, and unless we want to start throwing that in the mix too, an inline style is clearly preferable here.

src/components/admin-locations-detail-modules/doorways.tsx
guscost: Yeah, sort of weird.

src/components/admin-locations-detail-modules/doorways.tsx
guscost: The former. I don't have strong opinions, but do like that you can `export default function` where you can't with function expressions.

src/components/admin-locations-edit/index.tsx
guscost: 😭

src/components/admin-locations-edit/index.tsx
guscost: Fine, I'll add them

src/components/admin-locations-edit/index.tsx
guscost: 💯

src/components/spaces-snippets/index.tsx
guscost: Not yet, that is brand-new in the API. I'll add to @density/lib-api-types

src/components/admin-locations-edit/index.tsx
guscost: Grumble grumble, waste of vertical space

src/components/spaces-snippets/index.tsx
guscost: Added to @density/lib-api-types v3.0.3

PR #919, opened 2020-06-23
Update @density/lib-api-types and @density/ui

1egoman:
- Updated to new density ui and lib api types packages
- Migrated all hardcoded space representations to have new space fields
- Migrated to use space picker from density ui instead of the one in the dashboard
  - ~Only one small thing came up when I did this: there's both a checkbox implementation in density ui and a checkbox implementation in the dashboard, and they are styled differently. This means that any space picker that uses the multi-select version now uses the density ui checkbox instead of the dashboard checkbox.~ **This is no longer a problem, I migrated the dashboard's checkbox into density ui here: https://github.com/DensityCo/ui/pull/405**
  - ~Density UI checkbox design:~ <br/>Screen Shot 2020-06-23 at 4 14 40 PM
  - ~Dashboard checkbox design:~ <br />
Screen Shot 2020-06-23 at 4 16 31 PM


src/components/account-registration/index.tsx
```diff
@@ -115,20 +115,18 @@ export class AccountRegistration extends React.Component<any, any> {
 
             <div className={styles.accountRegistrationConsentContainer}>
               <label className={styles.accountRegistrationConsentLabel} htmlFor="account-registration-core-consent">
-                <input
-                  type="checkbox"
+                <Checkbox
```
1egoman, line 118: I ended up updating this checkbox instance while I was in here to the new one since it was trivial:
Screen Shot 2020-06-29 at 1 12 09 PM
Screen Shot 2020-06-29 at 1 12 13 PM
 
(before, it was just a raw `<input type="checkbox" />` 🤔 )

guscost: Any others of this method still floating around?

1egoman: Nope - there are two other bespoke checkbox implementations (one in the email digest modal and one on the user management page) but no other raw input checkboxes.

1egoman:
Yep, I have gone through both every space picker and every checkbox.

PR #948, opened 2020-07-24
Modify on prem login page to no longer use InputStack

1egoman:
I believe this was a regression that I accidentally introduced into the onprem build when merging https://github.com/DensityCo/dashboard/pull/943.


1egoman:
Yep, I'll be doing that on monday

PR #951, opened 2020-07-30
Density UI Dialogger migration

1egoman:
This completes the migration of the new `Dialogger` from density ui into the dashboard.

- This mostly consisted of swapping `showModal(dispatch, 'MODAL_CONFIRM', {` with `Dialogger.confirm({` (also alert and prompt) all over the place
- The new Dialogger returns a promise that eventually resolves to the result of the dialog after it closes, while the old api used a callback for this. In many cases, this actually made it simpler (since the old callback api was actually wrapped in a `new Promise()` call), but there were a lot of places I had to pull out what was in the callback and put it in a `.then` , or use async / await.
- I updated density ui to include implementations for `optimistic` and `initialText`, two parameters that the dialogger in the dashboard had before that the density ui version did now. (https://github.com/DensityCo/ui/pull/413)

## Dashboard Dialogger vs Density UI Dialogger API differences

There are a few api differences I want to call out:
1. As mentioned before, the new Dialogger returns a promise, while the old dialogger used a callback.
2. There was one api difference in the new prompt dialog: before, submitting a prompt that was empty / cancelling the prompt didn't cause the callback to fire, and now in the new implementation it resolves to be a nullable string. I prefer it to be this way, as I think it makes the abstraction a bit more useful, so I had to add checks to all the prompts to bail early if this condition was true.
3. Similar to the prompt, the old confirm dialog's callback was only called when the "confirm" button was clicked, and nothing happened when it was cancelled. Now, since the confirm dialog resolves a `boolean`, I had to bail out early everywhere the resolved value was false.

## Testing
I've interacted with each dialog, and made sure that it opened, and that clicking the "positive" button in each one did what was expected.


src/components/admin-locations-detail-modules/danger-zone.tsx
```diff
@@ -48,22 +47,24 @@ const ConnectedAdminLocationsDetailModulesDangerZone: React.FC = () => {
...
-        if (ok) {
-          Toaster.showToast({ text: 'Space deleted successfully' });
-        } else {
-          Toaster.showToast({ type: 'error', text: 'Error deleting space' });
-        }
-        // FIXME: this seems like a bad idea to have this just chilling in this callback
-        window.location.href = `#/admin/locations/${space.parent_id}`;
-      }
     });
+    if (!confirmed) {
```
1egoman, line 56: Example of "difference #3" - bailing out early when the negative button on a confirm box is clicked

src/components/analytics-control-bar/index.tsx
```diff
@@ -295,21 +289,22 @@ export const AnalyticsControlBarButtons: React.FunctionComponent<AnalyticsContro
...
+                  const newName = await Dialogger.prompt({
+                    title: `Rename '${reportName}'`,
+                    prompt: null,
+                    initialText: reportName,
                     placeholder: 'Enter a new name',
                     confirmText: 'Save',
-                    callback: onUpdateReportName,
                   });
+
+                  if (newName === null || newName.length === 0) { return; }
```
1egoman, line 306: Example of "difference #2": Bailing out early when cancel is clicked on a prompt, or the prompt is empty.

src/components/analytics-home/index.tsx
```diff
@@ -58,24 +53,29 @@ function MoreButton({reportName, onUpdateReportName, onDeleteReport}: MoreButton
...
+            Dialogger.confirm({
               title: `Delete Report`,
               prompt: `Are you sure you want to delete ${reportName}?`,
               confirmText: 'Delete',
-              callback: onDeleteReport,
+            }).then(confirmed => {
+              if (confirmed) {
+                onDeleteReport();
+              }
             });
```
1egoman, line 79: Example of "difference #1": `Dialogger.confirm` / `Dialogger.alert` / `Dialogger.prompt` now returns a promise.

src/components/analytics-control-bar/index.tsx
guscost: Odd that taking the affirmative action is even allowed when the prompt is empty, but OK.

1egoman:
Talked about gus's note with him synchronously, I'm going to make a ticket about this for the future.

PR #956, opened 2020-08-10
New environment switcher

1egoman:
- Migrate to the new environment switcher, used in a number of other internal apps
  - Still openable with the same old shortcut
  - Adds buttons to quickly pick an environment, the options being `Production`, `Staging`, `Local`, or `Custom`
  - Adds an option to toggle if the Auth0 login page is used when logging in via the `Local` and `Custom` environments
    - This makes local development easier.

-  Adds a "global" section to the bottom of the environment switcher
  - There's an option to toggle if a button is persistently shown on the navbar that allows the environment to be changed
    - (in other words, with this enabled, you don't have to type the keyboard shortcut all the time!)]
  - Copies over the "slow mode" toggle from the old environment switcher


src/components/login/index.tsx
```diff
@@ -28,15 +28,12 @@ import {
...
-import logoOktaO from '../../assets/images/logo-okta-o.png';
 
 export const  LOGIN = 'LOGIN',
-              OKTA = 'OKTA',
               STANDARD = 'STANDARD',
               FORGOT_PASSWORD = 'FORGOT_PASSWORD';
 
-export class OnPremLoginForm extends React.Component<any, any> {
+export class BasicLoginForm extends React.Component<any, any> {
   constructor(props) {
```
1egoman, line 37: NOTE: While I was in here, I noticed that all the auth0 login logic from the old login process no longer worked, and since this is page is only now used in local development and on prem (and both currently only use basic auth) this should be fine.

This multiple login mode thing has a lifespan though, and after the next on prem project is complete we should be able to standardize on the new common app login method everywhere pretty easily. If you want to know more details about this, we can talk in slack.

src/components/spaces-reports/index.tsx
```diff
@@ -9,7 +9,7 @@ import { Checkbox, DateRangePicker, Button, InputBox } from '@density/ui/src';
 import { CoreSpace } from '@density/lib-api-types/core-v2/spaces';
 import { REPORTS, ReportWrapper, ReportNoData, SingleReportLoading } from '@density/reports';
 
-import { getGoSlow } from '../environment-switcher';
+import { getGoSlow } from '../../helpers/unsafe-configure-app';
```
1egoman, line 12: I moved this helper alongside all the other environment switcher code, it fit better there.

src/helpers/unsafe-configure-app/index.ts
```diff
@@ -1,39 +1,66 @@
...
-// 2. Developer opens the environment switcher modal, changes an environment variable, then clicks
-// "ok". The `EnvironmentSwitcher` component's `onChange` is fired, which calls
-// `setServiceLocations`. The locations of all the services update.
-//
 export function configureClients() {
+  const state = EnvironmentStore.imperativelyGetValue();
+  const { coreApi, accountsApi } = (() => {
+    switch (state.configuration.type) {
+      case Environment.PRODUCTION:
+        return {
```
1egoman, line 16: This logic here does most of the adapting from the `EnvironmentStore` to the existing dashboard.

src/index.tsx
```diff
@@ -526,11 +522,6 @@ ReactDOM.render(
     <div>
       <App />
       <IntercomDensity />
-      <EnvironmentSwitcher
```
1egoman, line 529: Remove the old environment switcher from the root of the app - the new one is rendered in the navbar so that the button to open it is visible there.

src/components/environment-switcher/index.tsx
```diff
@@ -1,155 +1,493 @@
...
+import EnvironmentStore, {
+  DEFAULT_LOCAL_SETTINGS,
+  DEFAULT_CUSTOM_SETTINGS,
+  DEFAULT_SETTING_VALUES,
+} from '../../rx-stores/environment';
+
+// FIXME: the below two Any<FixInReview> types should refer to the same thing
+type EnvironmentDefinitionControls = {
+  controls: (conf: Any<FixInReview>, setConf: any) => React.ReactNode;
+  defaultValue: Any<FixInReview>;
```
1egoman, line 46: Help would be appreciated with the typings here. I want to do something like this:
```
type EnvironmentDefinitionControls<T> = {
  controls: (conf: T, setConf: T => void) => React.ReactNode;
  defaultValue: T;
}
```

And then later on be able to have the `T` inferred:
```
type Foo = { hello: 'world' } & EnvironmentDefinitionControls;
```

But that doesn't seem to work.

guscost: This seems to work:

type EnvironmentDefinitionControls<T> = {
  controls: (conf: T, setConf: (arg0: T) => void) => React.ReactNode;
  defaultValue: T;
}

type Foo = EnvironmentDefinitionControls<{ hello: 'world' }>;

guscost: Try to use inference more by describing the constants in a different way. Still one explicit `as` but it's looking pretty good.

tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "noErrorTruncation": true,
     "jsx": "preserve",
 
-    "lib": ["es2017", "dom", "esnext.array"],
+    "lib": ["es2019", "dom", "esnext.array"],
```
guscost, line 18: Waaaat

PR #970, opened 2020-08-14
Auto window height for PRD-1170

guscost:
Adds a hook to automatically (re)calculate the window height, and uses this hook to resize the analytics space picker for PRD-1170. Also lots of updates to consolidate other auto-sizing on a similar hook.


src/helpers/use-auto-size/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  windowHeight: number,
+};
+
+// Browser event listener hook
+export function useEventListener(eventName, handler, element = window){
+  const savedHandler = useRef((event) => undefined);
+  
+  useEffect(() => {
+    savedHandler.current = handler;
+  }, [handler]);
```
de1ux, line 19: thats a pretty neat feature 👨‍🍳

src/components/spaces/index.tsx
```diff
@@ -100,6 +101,7 @@ export default function Spaces() {
         <AppSidebar visible={true} width={spacesPage.navigationCollapsed ? 16 : 265}>
           {(spaces.status === RepositoryStatus.IDLE && spaces.data.size > 0) ? <div
             onClick={() => {
+              triggerResize(300);
```
guscost, line 104: This was an existing bug - collapsing the sidebar on the Spaces page did not resize elements on that page. This is the best way to do this in response to a layout change/UI action - all components that have useAutoSize will notice the triggered "resize" event and update accordingly.

src/helpers/use-auto-size/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// "Size" types are just a width and a height, for convenience
+export type Size = {
+  width: number,
+  height: number,
+};
+export type WindowSize = {
+  windowWidth: number,
+  windowHeight: number,
+};
```
guscost, line 11: Questionable whether these should be different types, but the destructuring ergonomics are really nice.

PR #994, opened 2020-08-20
Environment switcher types

guscost:


src/components/environment-switcher/index.tsx
```diff
@@ -415,7 +398,7 @@ const EnvironmentSwitcherModal: React.FunctionComponent<{}> = () => {
               environment.controls !== null ? (
                 <div className={styles.environmentChoiceControls}>
                   {environment.controls(
-                    state.workingConfiguration,
+                    state.workingConfiguration as Any<FixInRefactor>,
```
guscost, line 401: This function can't infer that it's one of the types which have config.

